PR #5, opened 2017-10-04
Triple-point and Sod test cases + allowing reduction to 1D runs.

homijan:
Very small changes are made in the computational kernel files, i.e. the option of SEGMENT geometry and corresponding simplification in the artificial viscosity tensor. In the main, the initialization of Sod shock tube and triple-point tests is implemented. Further, the information about the final state of hydro variables (max and min values) is printed. All new 1D/2D/3D runs are published in README (including graphics). In principle only the minor additions in laghos_solver.cpp are important, everything else can be avoided.


vladotomov:
Found the issue with -vis for the segment mesh.

homijan:
This issue was originally caused by mesh handling in MFEM, see https://github.com/mfem/mfem/pull/301.

tzanio (approved):

vladotomov:
Milan, do you need the Sod simulation on 2D / 3D meshes? I'm thinking to leave only the segment one. Also, do you mind removing the couts for min/max values of the functions?

homijan:
Vladimir, I agree that Sod makes sence in 1D (2D/3D was just a check
of completeness). The same for min/max, go ahead and remove these too.

2017-10-06 2:16 UTC+02:00, Vladimir Tomov <notifications@github.com>:

vladotomov (approved):

PR #12, opened 2018-02-23
Fixed a bug in the evaluation of the material gamma values.

vladotomov:
The gamma lookup was done by a FunctionCoefficient in physical coordinates, so once the materials moved, the gamma values became incorrect. This was affecting the behavior of the 3point test case.


laghos.cpp
```diff
@@ -215,11 +215,13 @@ int main(int argc, char *argv[])
    }
    int product = 1;
    for (int d = 0; d < dim; d++) { product *= nxyz[d]; }
+   /*
```
tzanio: Uncomment or delete.

tzanio (approved):

homijan (approved):
Thank you Vladimir, I thought that function coefficient was using fluid coordinates in Eval(). The GridFunctionCoefficient provides a sufficiently general and correct evaluation in Lag. coords.

PR #32, opened 2018-11-02
AMR version the 'amr' subdirectory [amr-in-master]

jakubcerveny:
This is just copied `amr-dev` into the `amr` directory. I still need to write a README. In short it will be something like:

* this is just a demo, only full assembly and Sedov are supported,
* the refinement/derefinement criteria are really stupid and only work for Sedov,
* the treatment of `h_0` needs some work,
* updating and inverting the mass matrices on mesh change is too expensive right now and there may be other things that are slowing the AMR version down - this also needs more work.

There is a companion pull request in MFEM: https://github.com/mfem/mfem/pull/660


tzanio (approved):

tzanio:
Thanks @jakubcerveny. Do you mind updating the `amr/README.md` file?

jakubcerveny:
I'll probably get to this no earlier than Thursday. (We've been moving over the weekend and things are still a bit hectic).

jakubcerveny:
I'm waiting for the last run for the verification table. Otherwise I think this is good.

vladotomov (approved):

v-dobrev (approved):

tzanio:
Thank you Jakub, this is a *really* cool demo.

I will merge this now so we stay on track for Laghos-1.2. Please add the data for the last run when you have it directly in `master`.

PR #33, opened 2018-11-02
Valgrind uninitialised warning fix

camierjs:
There was a conditional jump depending on uninitialised value for some `./laghos -m data/square01_quad.mesh -p 0 -tf 0.1` run.


vladotomov:
I'm fine merging this, but it doesn't fix the valgrind issue which shows up for -fa too. It's also weird that it doesn't show up in debug mode. I looked at SparseMatrix::AddSubMatrix, but all if statements seem fine to me.

PR #39, opened 2018-11-12
Adding Travis CI testing

### LINKED ISSUE #29: Add Travis testing
We should add a simple testing with Travis CI, at least for the official benchmark version.

tzanio:
Resolves #29

TODO
- [ ] Before merging, disable "Build Pushed Branches" in https://travis-ci.org/CEED/Laghos/settings
- [ ] Which version of MFEM to use (and cache): `laghos-1.0`, `master`, or `laghos-1.2`?
- [ ] Squash-merge


tzanio:
Should be squash-merged

v-dobrev (approved):

tzanio:
@vladotomov, any comments?

.travis.yml
```diff
@@ -0,0 +1,115 @@
...
+     else
+        echo "Reusing cached metis-4.0/";
+     fi;
+
+   # MFEM
+   - if [ ! -e mfem/libmfem.a ]; then
+        rm -rf mfem;
+        git clone --depth 1 https://github.com/mfem/mfem.git;
+        cd mfem;
+        git checkout laghos-1.0;
```
vladotomov: Should we make a new mfem tag?

.travis.yml
```diff
@@ -0,0 +1,115 @@
+language: cpp
+
+matrix:
+  include:
+    - os: linux
```
vladotomov, line 5: Can we also test on macOS?

.travis.yml
tzanio: We will create an MFEM tag for the laghos-2.0 release, but I'm actually leaning towards testing MFEM master in the Laghosn Travis (we can make Laghis changes in the future that require changes in MFEM).

What do you think?

.travis.yml
tzanio: Yes, but I was thinking of adding that in a follow-on PR.

.travis.yml
vladotomov: I prefer master too.

vladotomov (approved):

PR #44, opened 2019-03-20
Flush to zero tolerance (ftz_tol) option [dazftz]

camierjs:
A slowdown can be observed with the Sedov problem after some refinements, due to de-normals.
This PR adds a kernel after the RHS computation to filter them out if requested with a non-zero `fts-tol`.

This can be reproduced with the two following runs:
`./laghos -m data/square01_quad.mesh -ok 3 -ot 2 -s 7 -cgt 0 -cgm 50 -ms 4 -vs 1 -rs 2 -rp 4 -p 0`
`./laghos -m data/square01_quad.mesh -ok 3 -ot 2 -s 7 -cgt 0 -cgm 50 -ms 4 -vs 1 -rs 2 -rp 4 -p 1`


tzanio (approved):
Nice catch, Jean-Sylvain!

vladotomov (approved):

tzanio:
What values of `fts-tol` are sufficient to fix the slowdown in the above examples?

fischer1:
I would never run in practice with denorm turned on.


Would anyone???



________________________________

v-dobrev (approved):

PR #46, opened 2019-05-06
only include header if using openmpi

cwsmith:
This PR allows compilation with MPI implementations that are not OpenMPI.


tzanio (approved):

camierjs (approved):

camierjs:
Incorporated, @cwsmith, thanks!

PR #48, opened 2019-07-02
AMD HIP support in Laghos (Native HIP and RAJA)

noelchalmers:
This PR adds a HIP-enabled version of Laghos as well as updates the RAJA version of Laghos to use RAJA v0.8 and adds a RAJA+HIP compilation option.


hip/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Verification of Results
+
+To make sure the results are correct, we tabulate reference final iterations
+(`step`), time steps (`dt`) and energies (`|e|`) for the runs listed below:
+
+1. `mpirun -np 4 laghos -p 0 -m ../data/square01_quad.mesh -rs 3 -tf 0.75 -pa`
+2. `mpirun -np 4 laghos -p 0 -m ../data/cube01_hex.mesh -rs 1 -tf 0.75 -pa`
+3. `mpirun -np 4 laghos -p 1 -m ../data/square01_quad.mesh -rs 3 -tf 0.8 -pa -cfl 0.05`
+4. `mpirun -np 4 laghos -p 1 -m ../data/cube01_hex.mesh -rs 2 -tf 0.6 -pa -cfl 0.08`
```
vladotomov, line 105: Why do you change the CFL?

hip/README.md
```diff
@@ -0,0 +1,130 @@
...
+[capable exascale ecosystem](https://exascaleproject.org/what-is-exascale),
+including software, applications, hardware, advanced system engineering and early
+testbed platforms, in support of the nation’s exascale computing imperative.
+
+## Differences with the official benchmark version
+
+The HIP version differs from the official benchmark version of Laghos (in the
+top-level directory) in the following ways:
+
+1. Only problems 0 and 1 are defined
```
vladotomov, line 40: Why only 0 and 1?

noelchalmers:
@vladotomov The README in the hip directory is analogous to the README in the cuda directory, as overall contents of the folders themselves are analogous. I assumed runtime options and problems would be identical, but are these options in the cuda/hip READMEs incorrect?

hip/README.md
camierjs: It is as in all of the subdirectories there: the source `cuda` one has the same limitations.

hip/README.md
camierjs: Same thing, it is due to the first `cuda` subdir, where I changed these values.
Now that the MFEM 4.0 port is ready, we should be able to get the same results.

camierjs:
Tested on our server, works fine! Thanks @noelchalmers !

PR #59, opened 2020-06-12
Change code to output files in the run directory

kevinhkhuynh:
Also made minor changes to makefile so that the rom-dev branch can work correctly when cloned to run a baseline during regression testing

### DIFF
--- a/laghos_rom.cpp
+++ b/laghos_rom.cpp
@@ -123,12 +123,15 @@ void ROM_Sampler::Finalize(const double t, const double dt, Vector const& S)
     {
         cout << "X basis summary output" << endl;
         BasisGeneratorFinalSummary(generator_X, energyFraction);
+        PrintSingularValues(rank, "X", generator_X, energyFraction);
 
         cout << "V basis summary output" << endl;
         BasisGeneratorFinalSummary(generator_V, energyFraction);
+        PrintSingularValues(rank, "V", generator_V, energyFraction);
 
         cout << "E basis summary output" << endl;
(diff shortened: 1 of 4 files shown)


chldkdtn (approved):
Looks good to me!

laghos_rom.cpp
```diff
@@ -1142,6 +1145,25 @@ void ROM_Operator::Mult(const Vector &x, Vector &y) const
     }
 }
 
+void PrintSingularValues(const int rank, const std::string& name, CAROM::SVDBasisGenerator* bg, const double energyFraction)
```
dylan-copeland: `energyFraction` is not used and can be removed.

makefile
```diff
@@ -182,3 +183,6 @@ style:
 	@if ! $(ASTYLE) $(FORMAT_FILES) | grep Formatted; then\
 	   echo "No source files were changed.";\
 	fi
+
+$(shell mkdir -p run)
+$(shell mkdir -p run/ROMsol)
```
dylan-copeland, line 192: It would also be convenient to have `make clean` or `make cleanrun` do `rm -rf run`. This seems necessary when doing multiple runs, since the files from two runs can get mixed up.

dylan-copeland (changes requested):
This PR looks good, and I just have a couple of suggestions.

makefile
kevinhkhuynh: added!

dylan-copeland (approved):

PR #60, opened 2020-06-16
Fix restore

chldkdtn:
"fix_resotre" fixes the restore phase for the time windowing approach

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -710,13 +710,16 @@ int main(int argc, char *argv[])
     int steps = 0;
     BlockVector S_old(S);
 
+    StopWatch samplerTimer;
     int rom_window = 0;
     ROM_Sampler *sampler = NULL;
     if (rom_offline)
     {
         if (dtc > 0.0) dt = dtc;
+        samplerTimer.Start();
         sampler = new ROM_Sampler(myid, &H1FESpace, &L2FESpace, usingWindows ? twep[0] : t_final, dt, S, rom_staticSVD, rom_offsetX0, rom_energyFraction, rom_window);
         sampler->SampleSolution(0, 0, S);
+        samplerTimer.Stop();
     }
 
     ROM_Basis *basis = NULL;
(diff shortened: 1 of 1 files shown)


laghos.cpp
```diff
@@ -757,17 +757,47 @@ int main(int argc, char *argv[])
         // -restore phase
         // No need to specify t_final because the loop in -restore phase is determined by the files in ROMsol folder.
         // When -romhr or --romhr are used in -online phase, then -restore phase needs to be called to project rom solution back to FOM size
+        std::ifstream infile_tw_steps("run/tw_steps");
+        int nb_step;
+        nb_step = 1000000;
```
dylan-copeland: Why does this need to be initialized to a large number? Is it checked later whether this is overwritten to a smaller number? Also, the initialization and declaration can be on one line.

laghos.cpp
```diff
@@ -757,17 +757,47 @@ int main(int argc, char *argv[])
...
+        cout << "nb_step: " << nb_step << "\n";
```
dylan-copeland: Why print here? Isn't this set later, so that it should be printed later? The same goes for rom_dim* in the next 3 lines.

laghos.cpp
chldkdtn: Thank you for pointing it out, Dylan. Initializing nb_step to zero is completely fine because ti starts with 1. I have changed it.

laghos.cpp
chldkdtn: These prints were there for my debugging purpose. I forgot to remove them. Now, I removed them. Thank you for pointing it out.

chldkdtn:
I have accidentally included "fair_timing" branch to this PR. Sorry, guys. Basically, I am subtracting time of sampling from timeLoopTimer in the offline phase and subtracting the time of building hyper-reduced operator from timeLoopTimer in the online phase, especially for time windowing approach.

dylan-copeland (approved):

PR #61, opened 2020-06-16
ROM for RK2Avg solver

dylan-copeland:
This PR implements ROM for the RK2Avg solver.

### DIFF
--- a/laghos_rom.cpp
+++ b/laghos_rom.cpp
@@ -35,6 +35,10 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const& 
             generator_X->takeSample(X.GetData(), t, dt);
             generator_X->computeNextSampleTime(X.GetData(), dXdt.GetData(), t);
         }
+
+        // Without this check, libROM may use multiple time intervals, and without appropriate implementation
+        // the basis will be from just one interval, resulting in large errors and difficulty in debugging.
+        MFEM_VERIFY(generator_X->getNumBasisTimeIntervals() == 1, "Only 1 basis time interval allowed");
     }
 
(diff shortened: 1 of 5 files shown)


laghos_rom.cpp
```diff
@@ -48,6 +52,10 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const&
 
         generator_V->takeSample(V.GetData(), t, dt);
         generator_V->computeNextSampleTime(V.GetData(), dVdt.GetData(), t);
+
+        // Without this check, libROM may use multiple time intervals, and without appropriate implementation
+        // the basis will be from just one interval, resulting in large errors and difficulty in debugging.
+        MFEM_VERIFY(generator_X->getNumBasisTimeIntervals() == 1, "Only 1 basis time interval allowed");
```
chldkdtn, line 58: Although this is problematic for this particular simulation, this capability of libROM can be useful for dividing time window by the number of snapshots collected.

chldkdtn (commented):
Shouldn't we make "max_model_dim" in laghos_rom.hpp file to be available as a user option to avoid compilation level change?

laghos_rom.cpp
dylan-copeland: I agree, but in order to use that capability there needs to be implementation in Laghos, which is not done. In the future we may want to use multiple libROM time intervals, but until that is supported in Laghos we need to warn the user that the basis comes from just one interval and may be inaccurate if misused for the entire simulation.

chldkdtn: Sounds good, Dylan. I will approve and merge it to rom-dev. Kevin is not supposed to work on ROM this week.

chldkdtn (approved):

PR #63, opened 2020-06-19
twpTemp.csv is automaticaaly generated in the offline phase for the t…

chldkdtn:
If many time windows are required, it is a pain to generate twp.csv file manually. This PR will make twpTemp.csv file automatically during the offline phase, based on the cutoff values from singular value decomposition. As a default, the number of sample points for position and energy is doubled from the number of position and energy basis size, respectively, while the number of sample points for velocity is 20x of the number of velocity basis size.

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -967,9 +967,11 @@ int main(int argc, char *argv[])
                 if (usingWindows && t >= twep[rom_window] && rom_window < numWindows-1)
                 {
                     sampler->Finalize(t, last_dt, S, cutoff);
-                    outfile_twp << twep[rom_window] << ", ";
-                    outfile_twp << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2] << ", ";
-                    outfile_twp << 2*cutoff[0] << ", " << 20*cutoff[1] << ", " << 2*cutoff[2] << "\n";
+                    if (myid == 0) {
+                        outfile_twp << twep[rom_window] << ", ";
+                        outfile_twp << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2] << ", ";
+                        outfile_twp << 2*cutoff[0] << ", " << 20*cutoff[1] << ", " << 2*cutoff[2] << "\n";
+                    }
                     delete sampler;
 
                     rom_window++;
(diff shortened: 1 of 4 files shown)


kevinhkhuynh:
can the print statement be moved into laghos_rom.cpp under another function, similar to PrintSingularValues, and have it be called by ROM_Sampler::Finalize? and the cutoff variable can also be moved into ROM_Sampler::Finalize? unless it is planned to have cutoff as a user-defined option from the command line.

laghos.cpp
```diff
@@ -1128,9 +1130,11 @@ int main(int argc, char *argv[])
     {
         samplerTimer.Start();
         sampler->Finalize(t, dt, S, cutoff);
-        outfile_twp << twep[rom_window] << ", ";
-        outfile_twp << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2] << ", ";
-        outfile_twp << 2*cutoff[0] << ", " << 20*cutoff[1] << ", " << 2*cutoff[2] << "\n";
+        if (myid == 0) {
```
kevinhkhuynh: I see that outfile_twp is only created if we are using windows, but this if statement doesn't have a check for using windows. Should there be a check here for using windows?

chldkdtn: > can the print statement be moved into laghos_rom.cpp under another function, similar to PrintSingularValues, and have it be called by ROM_Sampler::Finalize? and the cutoff variable can also be moved into ROM_Sampler::Finalize? unless it is planned to have cutoff as a user-defined option from the command line.

I certainly see that this suggestion is a good idea and also I think that cutoff variables will not be a user-defined option from the command line in the future. However, we might need to introduce the user-defined coefficients that are multiplied by cutoff values to specify the number of sample points (for now it is hard-coded to 2, 20, 2) because these numbers are problem-specific and we do not have any mathematical justification for these numbers unlike cutoff values which can be justified by the singular value ratios.  This actually raises question about the current format of twp.csv file because the number of samples depends on the corresponding cutoff values. Let's discuss on this.

laghos.cpp
```diff
@@ -287,7 +287,8 @@ int main(int argc, char *argv[])
         MFEM_VERIFY(numWindows > 0, "");
         if (rom_online || rom_restore)
         {
-            const int err = ReadTimeWindowParameters(numWindows, twpfile, twep, twparam, myid == 0);
+            double sFactor[]  = {2.0, 20.0, 2.0};
```
dylan-copeland: Should we add 3 input parameters for these factors, instead of hard-coding them? These values could be the defaults in case the user does not specify them.

kevinhkhuynh: good idea! added.

dylan-copeland (approved):

PR #65, opened 2020-06-24
Add regression test

kevinhkhuynh:
This is the regression test suite. To run:

./tests/runRegressionTest.sh

The results are outputted into tests/results. Please let me know if you would like anything changed or simplified about how the information is outputted. Also, if you would like any of the parameters of the 32 simulations changed.

Currently, for the 4 commands of a test (offline, online, romhr, restore), I do them in that order and if one fails, I skip the rest and move onto the next test script.


chldkdtn:
I get the following error when executing ./tests/runRegressionTest.sh:

while: Expression Syntax.
do: Command not found.

kevinhkhuynh:
> I get the following error when executing ./tests/runRegressionTest.sh:
> 
> while: Expression Syntax.
> do: Command not found.

That's strange. I'm not getting this error, it might be because we are using different shells. The very first line of runRegressionTests.sh is the while loop. It looks like this

while getopts ":i:e:n:" o; do

Can you just press newline after the ; and make it like this? I think this should fix the problem. If not, please let me know. If it works, I'll push it to the pull request.

while getopts ":i:e:n:" o;
do

chldkdtn:
> > I get the following error when executing ./tests/runRegressionTest.sh:
> > while: Expression Syntax.
> > do: Command not found.
> 
> That's strange. I'm not getting this error, it might be because we are using different shells. The very first line of runRegressionTests.sh is the while loop. It looks like this
> 
> while getopts ":i:e:n:" o; do
> 
> Can you just press newline after the ; and make it like this? I think this should fix the problem. If not, please let me know. If it works, I'll push it to the pull request.
> 
> while getopts ":i:e:n:" o;
> do

I think you are right. If I use bash, then tests run fine. My shell tcsh. In tcsh, I still get the following error:

while: Expression Syntax.

Is it possible to make this script as shell-indepenent?

kevinhkhuynh:
> > > I get the following error when executing ./tests/runRegressionTest.sh:
> > > while: Expression Syntax.
> > > do: Command not found.
> > 
> > 
> > That's strange. I'm not getting this error, it might be because we are using different shells. The very first line of runRegressionTests.sh is the while loop. It looks like this
> > while getopts ":i:e:n:" o; do
> > Can you just press newline after the ; and make it like this? I think this should fix the problem. If not, please let me know. If it works, I'll push it to the pull request.
> > while getopts ":i:e:n:" o;
> > do
> 
> I think you are right. If I use bash, then tests run fine. My shell tcsh. In tcsh, I still get the following error:
> 
> while: Expression Syntax.
> 
> Is it possible to make this script as shell-indepenent?

I added #!/bin/bash to the top of the file which ensures that the script is run using bash no matter what shell (tcsh) you are using in your console. Does this fix the problem in your shell?

chldkdtn:
> > > > I get the following error when executing ./tests/runRegressionTest.sh:
> > > > while: Expression Syntax.
> > > > do: Command not found.
> > > 
> > > 
> > > That's strange. I'm not getting this error, it might be because we are using different shells. The very first line of runRegressionTests.sh is the while loop. It looks like this
> > > while getopts ":i:e:n:" o; do
> > > Can you just press newline after the ; and make it like this? I think this should fix the problem. If not, please let me know. If it works, I'll push it to the pull request.
> > > while getopts ":i:e:n:" o;
> > > do
> > 
> > 
> > I think you are right. If I use bash, then tests run fine. My shell tcsh. In tcsh, I still get the following error:
> > while: Expression Syntax.
> > Is it possible to make this script as shell-indepenent?
> 
> I added #!/bin/bash to the top of the file which ensures that the script is run using bash no matter what shell (tcsh) you are using in your console. Does this fix the problem in your shell?

It works in my shell! Thanks, Kevin. By the way, I still see all the screen outputs during the regression tests. Can you put them in a log file so that we do not see those screen outputs during the regression tests. Instead, in the screen, I would like to see something like:

[test 1: sedov-time-window-offline] pass 
[test 2: sedov-time-window-online] fail: failure log file is written in xxx.log
.
.
.

Something like this. Will this be implemented in another separate PR?

No problem! The output has been suppressed to log files.

dylan-copeland:
I will not have time to review this PR this week, but you can merge it without my approval. I doubt I would have much input about testing anyway.

kevinhkhuynh:
> > > > > I get the following error when executing ./tests/runRegressionTest.sh:
> > > > > while: Expression Syntax.
> > > > > do: Command not found.
> > > > 
> > > > 
> > > > That's strange. I'm not getting this error, it might be because we are using different shells. The very first line of runRegressionTests.sh is the while loop. It looks like this
> > > > while getopts ":i:e:n:" o; do
> > > > Can you just press newline after the ; and make it like this? I think this should fix the problem. If not, please let me know. If it works, I'll push it to the pull request.
> > > > while getopts ":i:e:n:" o;
> > > > do
> > > 
> > > 
> > > I think you are right. If I use bash, then tests run fine. My shell tcsh. In tcsh, I still get the following error:
> > > while: Expression Syntax.
> > > Is it possible to make this script as shell-indepenent?
> > 
> > 
> > I added #!/bin/bash to the top of the file which ensures that the script is run using bash no matter what shell (tcsh) you are using in your console. Does this fix the problem in your shell?
> 
> It works in my shell! Thanks, Kevin. By the way, I still see all the screen outputs during the regression tests. Can you put them in a log file so that we do not see those screen outputs during the regression tests. Instead, in the screen, I would like to see something like:
> 
> [test 1: sedov-time-window-offline] pass
> [test 2: sedov-time-window-online] fail: failure log file is written in xxx.log
> .
> .
> .
> 
> Something like this. Will this be implemented in another separate PR?
> 
> No problem! The output has been suppressed to log files.

No problem! The output has been suppressed to log files.

chldkdtn:
Looks fantastic, Kevin! One last request, though. At the end of all the tests, can you print a summary of how many tests are passed and how many tests are failed, e.g., 

31 passed, 1 failed among 32 tests

kevinhkhuynh:
> Looks fantastic, Kevin! One last request, though. At the end of all the tests, can you print a summary of how many tests are passed and how many tests are failed, e.g.,
> 
> 31 passed, 1 failed among 32 tests

Added

kevinhkhuynh:
LGTM

chldkdtn (approved):

PR #68, opened 2020-07-01
Match time window changes online

dylan-copeland:
This PR resets the timestep `dt` so that online timesteps exactly match time window changes. This could result in an extremely small timestep at a window change, which is not checked. If this causes problems, it should be easy to notice the small `dt` in the output. Also, times are now read from the CSV file as doubles (previously as floats, causing some round-off in the window definitions). One test (Gresho) already showed that this change decreases the online ROM error significantly.

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -893,6 +893,9 @@ int main(int argc, char *argv[])
                 last_step = true;
             }
 
+            if (rom_online && usingWindows && (t + dt >= twep[rom_window]))
+                dt = twep[rom_window] - t;
+
             if (steps == max_tsteps) {
                 last_step = true;
             }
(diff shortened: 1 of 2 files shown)


laghos.cpp
```diff
@@ -893,6 +893,9 @@ int main(int argc, char *argv[])
                 last_step = true;
             }
 
+            if (rom_online && usingWindows && (t + dt >= twep[rom_window]))
+                dt = twep[rom_window] - t;
```
chldkdtn, line 907: When dt becomes very small due to this, it takes quite a while to recover a reasonable time step. In other words, the small time step affects the next time step, causing the overall simulation time to be long. Can we store previous dt so and after matching the end time of a window, use the previous dt again as dt? or something similar?

chldkdtn: I have pushed the one version that achieves bringing dt_old back after matching end time of each window. However, please modify as you see any necessary change.

chldkdtn (approved):

dylan-copeland (commented):
The last commit for `use_dt_old` looks good to me.

PR #70, opened 2020-07-07
ROM offset

dylan-copeland:
This PR enables offsetting the ROM spaces for X, V, and E by the initial state of the variables. To enable, use command line flag `-romos`. Initial testing showed that the error was significantly reduced, but the ROM basis sizes were larger, so it is unclear whether the offset or the larger bases or both are the cause for better accuracy.

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -757,6 +757,13 @@ int main(int argc, char *argv[])
     Vector romS, romS_old;
     ROM_Operator *romOper = NULL;
 
+    if (!usingWindows)
+    {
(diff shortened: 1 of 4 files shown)


chldkdtn (changes requested):
1) -romhr is not used (no hyperreduction), and the offsets are not read in that case.
2) -nsam{x,v,e} are not set, and the default factors are used only in the case of using windows.

chldkdtn:
LGTM

chldkdtn (approved):

PR #79, opened 2020-08-27
Fix regression tests

dylan-copeland:


dylan-copeland:
This PR fixes one broken test, but `sedov-blast-parameter-variation-online-parallel` is still failing (basis files are not found?). @kevinhkhuynh do you want to look into that test?

kevinhkhuynh:
> 
> 
> This PR fixes one broken test, but `sedov-blast-parameter-variation-online-parallel` is still failing (basis files are not found?). @kevinhkhuynh do you want to look into that test?

Hi, Dylan. Would you happen to know if these sequence of commands are correct? I just copied the serial version to parallel, but maybe I had to change something in the command. When I run the test normally, it fails on the last command due to basis files not existing.

srun -n 8 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -romsrhs -bef 1.0 -rpar 0
./merge -nset 1 -rhs
srun -n 8 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -bef 0.5 -writesol -visit
srun -n 8 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -online -rdimx 32 -rdimv 92 -rdime 26 -rdimfv 200 -rdimfe 60 -romhr -romos -sfacx 1 -sfacv 32 -sface 32 -soldiff -romrmass -romsrhs -bef 1.0

dylan-copeland:
@kevinhkhuynh It turns out ./merge is working only for serial tests. I will extend it to the parallel case.

dylan-copeland:
@kevinhkhuynh Thanks for changing the tests to work with sbatch. I have been running with `mxterm`, but sbatch seems more convenient. One issue is that the regression tests are taking over 30 minutes now, and the `pdebug` limit is 30 minutes, so I have been running one directory of tests at a time, e.g. `runRegressionTests.sh -i "sedov-blast"`. Can this be done with the new version using sbatch?

chldkdtn:
@kevinhkhuynh In addition to @dylan-copeland 's comments above, is it possible to run several tests at the same time after grabbing enough cores initially with sbatch?

dylan-copeland:
@kevinhkhuynh The failing test can be fixed by changing `./merge -nset 1 -rhs` to `srun -n 8 -p pdebug ./merge -nset 1 -rhs` when running manually, in `sedov-blast-parameter-variation.sh`. I am not sure how this should be changed, with your changes for sbatch. Could you please update this?

kevinhkhuynh:
@dylan-copeland @chldkdtn Yes, I'll make the changes. Also, sbatch currently uses "pbatch" but the type can be changed to "pdebug" at the top.

dylan-copeland:
@kevinhkhuynh With your sbatch changes, will `./tests/runRegressionTests.sh` on rztopaz use the login node for running the script but not for running the tests, i.e. the tests are run in batch mode rather than interactively on the login node?

kevinhkhuynh:
@dylan-copeland Using sbatch will work with using the option "-i" command. Additionally, everything, including the tests, will be run on the 1 node that is obtained by sbatch. The login node will not be used for anything. sbatch.log will contain most of the information that would normally have been outputted to the terminal. sbatch.log won't show "TEST: running" anymore, but after a test is run, will show "TEST: pass" or "TEST: fail." The individual tests running can still be viewed inside results/*.

kevinhkhuynh:
> 
> 
> @kevinhkhuynh In addition to @dylan-copeland 's comments above, is it possible to run several tests at the same time after grabbing enough cores initially with sbatch?

This can be done, but we will have to add a parameter to Laghos to output the data in any directory we wish. Otherwise, the different processors will overwrite each other when running the different simulations in the run directory. I will work on adding this.

dylan-copeland:
The sbatch change is a nice improvement. Now everything is passing except the two gresho tests where -romhr is not used (online without hyperreduction is not really supported in that case) and sedov-blast-parameter-variation-fom-parallel, which fails because I changed merge to output the singular values `sVal*` only on the root process and the test is trying to compare files for non-root ranks. This should be fine after merging to rom-dev.

PR #77, opened 2020-08-20
Add parametric ROM capability to physical time windowing RHS optimization

siuwuncheung:
This PR is a work in progress and is not ready to merge yet.

For physical time windowing, multiple offline computations are run to generate a snapshot set for each parameter and time window. For instance,  

`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -romsrhs -rhof 1.0 -rpar 0 -nwin 2 -tw tw2.csv`
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -romsrhs -rhof 2.0 -rpar 1 -nwin 2 -tw tw2.csv`
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -romsrhs -rhof 0.5 -rpar 2 -nwin 2 -tw tw2.csv`

Here, `rho` scaling factor `-rhof` is the varying parameter and `nwin` is the number of time windows. In the `run` directory, snapshot files `paramJ_varNT_snapshot.rank` are generated, where `J` is the parameter choice index (0-based), `N` is the variable name (`X`, `V`, `E`), `T` is the time window index (0-based), and `rank` is the MPI rank.

To generate the basis, running

`./merge -nset 3 -ef 0.9999 -rhs -nwin 2 -tw tw2.csv` 

will merge the snapshots for each variable in each time window, over all parameter indices, generate `basisNT.rank` and `sValNT.rank` files, and print out recommended basis sizes corresponding to the prescribed energy fraction.

For online simulation, comparing with the full-order solution 

`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -rhof 0.75 -writesol -visit`

run 

`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -online -romhr -romos -sfacx 1 -sfacv 32 -sface 32 -soldiff -romgs -romsrhs -rhof 0.75 -rpar 3 -nwin 2 -tw tw2.csv -twp twpTemp.csv`

PR #87, opened 2020-09-11
Remove redundant sampling when time windows are finalized

siuwuncheung:
This PR removes the redundant sample in the `ROM_Sampler::Finalize` call of the time windows. The same snapshot vector has been sampled in `ROM_Sampler::SampleSolution` call.


dylan-copeland:
@siuwuncheung This PR looks good to me, and the only thing left to verify is that it doesn't break anything. The regression tests are failing, which makes sense as the basis sizes are changed. Can you comment on whether you verified the differences in the tests are minor (e.g. just different basis sizes)? For example, you may be able to just comment out `check_fail` after `basisComparator` is called in `runRegressionTests.sh`, to eliminate failures due to basis differences. There could still be other small differences that need to be checked manually (e.g. compare ROM online solution errors that may be slightly different). If you describe your verification, it can help the reviewers.

chldkdtn:
@siuwuncheung , please resolve the conflict introduced due to the recent merge of PR #77 to the rom-dev.

siuwuncheung:
@dylan-copeland @chldkdtn It turns out many regression tests break. As the snapshot vectors are different, everything including singular values, singular vectors, time steps and the solution are different. By manually read the online stage log files one by one, the runs are all successfully complete, but the results are different from the baseline. In most of the tests, the differences are slight as expected, but there are few tests which yield very big difference.

dylan-copeland (approved):
This change seems fine, and I am relying on your verification @siuwuncheung that the regression tests had acceptable small changes.

PR #88, opened 2020-09-15
Using the initial condition as an offset for all the time windows

siuwuncheung:
This PR adds the option of using different styles of offsets in time window and implemented the choice of passing the initial state as offset in all windows. The user command option `-rostype` controls `romOptions.romOffsetType` (default 0), in both offline and online stage. 

0 -> save and load offset; interpolate in parametric case (default)
1 -> use initial state as offset
2 -> use solution from previous window as offset

In this PR, the implementation of option 0 with parametric case is not complete, in which the results produced will be same as option 2.

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -737,7 +737,7 @@ int main(int argc, char *argv[])
     romOptions.window = 0;
     romOptions.FOMoper = &oper;
     romOptions.parameterID = rom_paramID;
-    romOptions.online = rom_online;
+    romOptions.restore = rom_restore;
 
     // Perform time-integration (looping over the time iterations, ti, with a
     // time-step dt). The object oper is of type LagrangianHydroOperator that
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -838,7 +838,10 @@ int main(int argc, char *argv[])
         romS.SetSize(romOptions.dimX + romOptions.dimV + romOptions.dimE);
         basis->ProjectFOMtoROM(S, romS);
 
-        cout << "Offset Style: " << offsetType << endl;
+        if (myid == 0)
+        {
+            cout << "Offset Style: " << offsetType << endl;
+        }
         cout << myid << ": initial romS norm " << romS.Norml2() << endl;
 
         romOper = new ROM_Operator(romOptions, basis, rho_coeff, mat_coeff, order_e, source, visc, cfl, p_assembly,
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -737,6 +737,7 @@ int main(int argc, char *argv[])
     romOptions.window = 0;
     romOptions.FOMoper = &oper;
     romOptions.parameterID = rom_paramID;
+    romOptions.online = rom_online;
 
     // Perform time-integration (looping over the time iterations, ti, with a
     // time-step dt). The object oper is of type LagrangianHydroOperator that
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -284,6 +284,7 @@ int main(int argc, char *argv[])
     args.AddOption(&rom_paramID, "-rpar", "--romparam", "ROM offline parameter index.");
     args.AddOption(&romOptions.paramOffset, "-rparos", "--romparamoffset", "-no-rparos", "--no-romparamoffset",
                    "Enable or disable parametric offset.");
+    args.AddOption(&romOptions.offsetType, "-rostype", "--romoffsettype", "Offset type for initializing ROM windows.");
     args.Parse();
     if (!args.Good())
     {
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -294,7 +294,11 @@ int main(int argc, char *argv[])
     args.AddOption(&rom_paramID, "-rpar", "--romparam", "ROM offline parameter index.");
     args.AddOption(&romOptions.paramOffset, "-rparos", "--romparamoffset", "-no-rparos", "--no-romparamoffset",
                    "Enable or disable parametric offset.");
-    args.AddOption(&romOptions.offsetType, "-rostype", "--romoffsettype", "Offset type for initializing ROM windows.");
+    args.AddOption(&romOptions.offsetType, "-rostype", "--romoffsettype",
+                   "Offset type for initializing ROM windows.\n\t"
+                   "0 -> save and load offset;\n\t"
+                   "1 -> use initial state as offset;\n\t"
+                   "2 -> use solution from previous window as offset");
     args.Parse();
     if (!args.Good())
     {
@@ -311,9 +315,9 @@ int main(int argc, char *argv[])
         const char path_delim = '/';
         std::string::size_type pos = 0;
         do {
-          pos = outputPath.find(path_delim, pos+1);
-          std::string subdir = outputPath.substr(0, pos);
-          mkdir(subdir.c_str(), 0777);
+            pos = outputPath.find(path_delim, pos+1);
+            std::string subdir = outputPath.substr(0, pos);
+            mkdir(subdir.c_str(), 0777);
         }
         while (pos != std::string::npos);
         mkdir((outputPath + "/ROMoffset").c_str(), 0777);
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -61,6 +61,7 @@
 #include "laghos_rom.hpp"
 #include "laghos_utils.hpp"
 #include <fstream>
+#include <map>
 
 #ifndef _WIN32
 #include <sys/stat.h>  // mkdir
@@ -184,6 +185,7 @@ int main(int argc, char *argv[])
     double rhoFactor = 1.0;
     int rom_paramID = -1;
     const char *normtype_char = "l2";
+    const char *offsetType = "load";
     Array<double> twep;
     Array2D<int> twparam;
     ROM_Options romOptions;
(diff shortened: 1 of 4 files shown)


laghos_rom.cpp
```diff
@@ -375,48 +375,100 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, Vector const& S, MPI_Comm comm_)
...
+                {
+                    gfL2[i] = S[2*H1size+i];
+                }
+                gfL2.GetTrueDofs(E);
+                for (int i=0; i<tL2size; ++i)
+                {
+                    (*initE)(i) = E[i];
+                }
+
+                initX->write("run/ROMoffset/initX0");
```
dylan-copeland: @siuwuncheung Why does `ROM_Basis` write the offset vectors to file? It seems it would only be necessary for the restore phase, after the online phase. For `offsetType == 2`, the offset is the initial state for this window, so the restore phase would need the offsets written to file for each window, right? If so, why is the window index not in the filename? For `offsetType == 1`, the offset is the initial state at `t=0`, which should have been written to file in the offline phase by `ROM_Sampler`, right?

siuwuncheung: @dylan-copeland Thanks for your review. Yes you're right about the case `offsetType == 2`, in which writing the offset solution is not necessary in the online phase but the restore phase. As you have suggested, the filename should include the window index, which is done by the latest commit. 

For `offsetType == 1`, the offset is written to file in the offline phase by `ROM_Sampler` for the non-parametric case. But for the parametric case, the offset is dependent on the parameter and has to be computed and saved at the first time window. There should be an upcoming fix for the restore phase in `offsetType == 1` too, as it just has one offset file even with multiple time windows.

dylan-copeland: @siuwuncheung Thanks for the update and explanation. It looks like the restore phase is not implemented correctly, as it should read the offset rather than writing it, but are you saying that will be fixed in a future commit or PR?

siuwuncheung: @dylan-copeland Sorry for the confusion caused. What I meant was the changes done in the PR so far was only about the online phase. For `offsetType == 2`, the offset is written to file in the online phase, which is used for future read in the restore phase. The restore phase has not be taken care yet, but shall be done shortly in this PR. On the other hand, for `offsetType == 1`, there needs to be a further modification, as the offset file is only written in the online phase when `input.window == 0`.

dylan-copeland: @siuwuncheung Ok, please let me know when the PR is ready for review.

siuwuncheung: @dylan-copeland The PR is ready for review now, with the restore phase is revised after commit 35f8ba1.

dylan-copeland:
@siuwuncheung This is the first time I have seen a "force-push". Was this intentional, and if so, why was it necessary?

siuwuncheung:
> @siuwuncheung This is the first time I have seen a "force-push". Was this intentional, and if so, why was it necessary?

@dylan-copeland I made a mistake in the commit 35f8ba1 that breaks the code that works earlier. I wanted to restore the earlier commits by force push. Maybe I should just undo the change and push a new commit. I think the code is working now with commit c77ed60 and I am checking with the regression tests again.

dylan-copeland:
@siuwuncheung @kevinhkhuynh @chldkdtn  Now `git pull` tries to merge the remote version with my local version, which have conflicts, and I'm not sure how to get out of it. Does someone know how to fix this?

siuwuncheung:
> e the remote version with my local version, which have conflicts, and I'm not sure how to get out of it. Does someone know how to fix t

Would `git reset --hard origin/romoffsetoption` work?

dylan-copeland:
@siuwuncheung I am sure there is a good way to fix this in git, but I am not a git expert, and I think the easiest solution considering there are only 97 new lines in this PR, would be to copy the files to a new directory on your machine, make a new branch, copy them over, and work in that new branch. You can try the hard reset, or other ideas, but if those do not work we have the simple approach of making a new branch/PR.

kevinhkhuynh:
> 
> 
> @siuwuncheung @kevinhkhuynh @chldkdtn Now `git pull` tries to merge the remote version with my local version, which have conflicts, and I'm not sure how to get out of it. Does someone know how to fix this?

It seems that when I just checked out the branch right now, there were no problem. 

I think this can fix your problems, if it isn't working. Though Tony can switch to a new branch if he would like.

1. git merge --abort
2. git fetch --all
3. git reset --hard origin/romoffsetoption

laghos_rom.hpp
```diff
@@ -65,6 +66,7 @@ struct ROM_Options
     bool GramSchmidt = false; // whether to use Gram-Schmidt with respect to mass matrices
     bool RK2AvgSolver = false; // true if RK2Avg solver is used for time integration
     bool paramOffset = false; // used for determining offset options in the online stage, depending on parametric ROM or non-parametric
+    int offsetType = 0; // 0 -> save and load offset; 1 -> use initial state as offset; 2 -> use solution from previous window as offset
```
kevinhkhuynh: Is it possible to switch to an int enum for readability, instead of 0, 1, and 2, so the offsetType is embedded in the variable name?

dylan-copeland:
@kevinhkhuynh Thanks, your suggestion worked for me. Now there is no conflict with my local branch.

laghos_rom.cpp
```diff
@@ -375,9 +375,28 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, Vector const& S, MPI_Comm comm_)
         initV = new CAROM::Vector(tH1size, true);
         initE = new CAROM::Vector(tL2size, true);
 
-        //TODO: Tony param init after PR 77
-        if (input.paramOffset)
+        // std::string path_init = (parameterID >= 0) ? "run/ROMoffset/param" + std::to_string(parameterID) + "_init" : "run/ROMoffset/init"; // TODO: Tony PR77
+        std::string path_init = "run/ROMoffset/init";
```
kevinhkhuynh: I pushed some changes to rom-dev that will affect this line. Can you change it to 

std::string path_init = basename + "/ROMoffset/init";

laghos_rom.hpp
```diff
@@ -183,30 +186,45 @@ class ROM_Sampler
...
             Xdiff.SetSize(tH1size);
             Ediff.SetSize(tL2size);
 
-            for (int i=0; i<tH1size; ++i)
-            {
-                (*initX)(i) = X[i];
-            }
-            initX->write(path_init + "X" + std::to_string(window));
+            //std::string path_init = (parameterID >= 0) ? "run/ROMoffset/param" + std::to_string(parameterID) + "_init" : "run/ROMoffset/init"; // TODO: Tony PR77
+            std::string path_init = "run/ROMoffset/init";
```
kevinhkhuynh: I pushed some changes to rom-dev that will affect this line. Can you change it to 

std::string path_init = basename + "/ROMoffset/init";

laghos_rom.hpp
siuwuncheung: @kevinhkhuynh Thanks Kevin. I will look into this switch.

dylan-copeland: I agree, an enum would make the code clearer.

chldkdtn (approved):

dylan-copeland:
@siuwuncheung Could you please add the enum for the options?

laghos_rom.hpp
```diff
@@ -66,7 +86,7 @@ struct ROM_Options
     bool GramSchmidt = false; // whether to use Gram-Schmidt with respect to mass matrices
     bool RK2AvgSolver = false; // true if RK2Avg solver is used for time integration
     bool paramOffset = false; // used for determining offset options in the online stage, depending on parametric ROM or non-parametric
-    int offsetType = 0; // 0 -> save and load offset; 1 -> use initial state as offset; 2 -> use solution from previous window as offset
+    offsetStyle offsetType = saveLoadOffset; // types of offset in time windows
```
dylan-copeland, line 89: It looks like this is not used and can be removed.

laghos.cpp
```diff
@@ -838,6 +838,7 @@ int main(int argc, char *argv[])
         romS.SetSize(romOptions.dimX + romOptions.dimV + romOptions.dimE);
         basis->ProjectFOMtoROM(S, romS);
 
+        cout << "Offset Style: " << offsetType << endl;
```
dylan-copeland: `if (myid == 0)` so that this is not printed by every rank.

dylan-copeland (commented):
Comments like the following can be updated with the new enum names, to eliminate the integer values: `// Offline phase rostype 1 time window > 0: Read the initial state`

chldkdtn:
@siuwuncheung it has conflict with the current rom-dev branch right now. Would you resolve it?

siuwuncheung:
> @siuwuncheung it has conflict with the current rom-dev branch right now. Would you resolve it?



> @siuwuncheung it has conflict with the current rom-dev branch right now. Would you resolve it?

@chldkdtn Resolved, thanks!

siuwuncheung:
> @siuwuncheung it has conflict with the current rom-dev branch right now. Would you resolve it?

Resolved, thanks!

siuwuncheung:
> Comments like the following can be updated with the new enum names, to eliminate the integer values: `// Offline phase rostype 1 time window > 0: Read the initial state`

Comments updated, thanks!

laghos.cpp
```diff
@@ -61,6 +61,7 @@
 #include "laghos_rom.hpp"
 #include "laghos_utils.hpp"
 #include <fstream>
+#include <map>
```
dylan-copeland: Can this be removed?

laghos_rom.cpp
```diff
@@ -353,9 +353,22 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, Vector const& S, MPI_Comm comm_,
...
+            // Read offsets in the restore phase or in the online phase of non-parametric save-and-load mode
+            initX->read(path_init + "X" + std::to_string(input.window));
+            initV->read(path_init + "V" + std::to_string(input.window));
+            initE->read(path_init + "E" + std::to_string(input.window));
+
+            cout << "Read init vectors X, V, E with norms " << initX->norm() << ", " << initV->norm() << ", " << initE->norm() << endl;
+        }
+        else if (input.offsetType == saveLoadOffset && input.paramOffset)
         {
+            // TODO: Tony interpolation PR 77
```
dylan-copeland, line 370: Should this TODO still be here?

siuwuncheung: Please ignore those TODOs. They are to be done in another PR and just left there to remind me.

laghos.cpp
siuwuncheung: Yes. The `unordered_map` call still works.

dylan-copeland (approved):

kevinhkhuynh (approved):

PR #89, opened 2020-09-15
librom basis generator options libROM #PR44

kevinhkhuynh:
Unfortunately, brace initialization is only allowed when the struct has default values in C++ 14, and designated initializer lists only exist in C++ 20.

Verified, all tests passing.

### DIFF
--- a/laghos_rom.hpp
+++ b/laghos_rom.hpp
@@ -95,78 +95,109 @@ public:
 
         if (input.staticSVD)
         {
-            generator_X = new CAROM::StaticSVDBasisGenerator(tH1size, max_model_dim,
+            generator_X = new CAROM::StaticSVDBasisGenerator(
+                    GenerateStaticSVDOptions(
+                      tH1size,
+                      max_model_dim
+                    ),
                     BasisFileName(VariableName::X, window, parameterID));
-            generator_V = new CAROM::StaticSVDBasisGenerator(tH1size, max_model_dim,
+            generator_V = new CAROM::StaticSVDBasisGenerator(
+                    GenerateStaticSVDOptions(
+                      tH1size,
+                      max_model_dim
+                    ),
                     BasisFileName(VariableName::V, window, parameterID));
-            generator_E = new CAROM::StaticSVDBasisGenerator(tL2size, max_model_dim,
+            generator_E = new CAROM::StaticSVDBasisGenerator(
+                    GenerateStaticSVDOptions(
+                      tL2size,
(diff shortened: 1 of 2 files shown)


laghos_rom.hpp
```diff
@@ -95,78 +95,109 @@ class ROM_Sampler
...
                         BasisFileName(VariableName::Fe, window, parameterID));
             }
         }
         else
         {
-            generator_X = new CAROM::IncrementalSVDBasisGenerator(tH1size,
-                    model_linearity_tol,
-                    false,
-                    true,
-                    max_model_dim,
```
kevinhkhuynh, line 119: @chldkdtn Do we ever expect generator_X, generator_V, generator_E, generator_Fv, and generator_Fe to have different parameter values for the SVDBasisGenerators besides the dim parameter?

laghos_rom.hpp
```diff
@@ -214,6 +245,19 @@ class ROM_Sampler
...
+    CAROM::StaticSVDOptions GenerateStaticSVDOptions(int dim, int samples_per_time_interval)
+    {
+      CAROM::StaticSVDOptions static_svd_options(dim, samples_per_time_interval);
+      return static_svd_options;
+    }
+
+    CAROM::IncrementalSVDOptions GenerateIncrementalSVDOptions(int dim, double linearity_tol, bool fast_update, int max_basis_dimension, double initial_dt, int samples_per_time_interval, double sampling_tol, double max_time_between_samples)
+    {
+      CAROM::IncrementalSVDOptions inc_svd_options(dim, linearity_tol, max_basis_dimension, initial_dt, samples_per_time_interval, sampling_tol, max_time_between_samples);
+      inc_svd_options.fast_update = fast_update;
```
kevinhkhuynh, line 257: This is a bit verbose compared to before, but it seems it's the only way we can get around potential backwards compatibility issues since partial brace initialization isn't allowed in C++11 when the struct has default values.

laghos_rom.hpp
dylan-copeland: I don't think so. Also, X, V, Fv should have the same dim, as should E and Fe, so the same options could be reused.

dylan-copeland (approved):

laghos_rom.hpp
chldkdtn: I agree with Dylan

chldkdtn (approved):

siuwuncheung (approved):

PR #91, opened 2020-09-17
filter small sv

kevinhkhuynh:
Filter small SV

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -173,6 +173,9 @@ int main(int argc, char *argv[])
     double sFactorX = 2.0;
     double sFactorV = 20.0;
     double sFactorE = 2.0;
+    double model_linearity_tol = 1.e-7;
+    double model_singular_value_tol = 1.e-14;
+    double model_sampling_tol = 1.e-7;
     int numWindows = 0;
     int windowNumSamples = 0;
     int windowOverlapSamples = 0;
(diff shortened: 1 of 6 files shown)


kevinhkhuynh:
What's a good value for this SV tolerance, or do we even want to use this? Setting SV to 1e-14 passes all tests, and doesn't filter anything as it's too small.

chldkdtn:
> What's a good value for this SV tolerance, or do we even want to use this? Setting SV to 1e-14 passes all tests, and doesn't filter anything as it's too small.

@kevinhkhuynh Setting SV tolerance to 1e-14 for now is good to keep the current tests valid. Would you add one new test though that has a larger SV tolerate, such as 1e-6?

kevinhkhuynh:
Ready for review.

chldkdtn (approved):
Looks good to me!

laghos_rom.hpp
```diff
@@ -109,9 +114,10 @@ class ROM_Sampler
     {
         const int window = input.window;
 
-        // TODO: read the following parameters from input?
-        double model_linearity_tol = 1.e-7;
-        double model_sampling_tol = 1.e-7;
+        if (sampleF)
```
siuwuncheung: This is no longer necessary. @dylan-copeland Is that right?

laghos.cpp
```diff
@@ -173,6 +173,9 @@ int main(int argc, char *argv[])
     double sFactorX = 2.0;
     double sFactorV = 20.0;
     double sFactorE = 2.0;
+    double model_linearity_tol = 1.e-7;
```
dylan-copeland: These 3 variables are not used and should be removed.

laghos_rom.hpp
```diff
@@ -78,6 +78,11 @@ struct ROM_Options
     int dimFv = -1;
     int dimFe = -1;
 
+    // Incremental SVD options
+    double model_linearity_tol = 1.e-7;
```
dylan-copeland: I think these are the historical names of these variables. Maybe it is time to update them to something more descriptive. First, `model` is not helpful. Second, if they are only used for incremental SVD, how about something like `model_*` --> `incSVD_*`?

dylan-copeland (approved):

chldkdtn:
@siuwuncheung Would you review this when you get a chance?

laghos_rom.hpp
kevinhkhuynh: removed

siuwuncheung (approved):

PR #93, opened 2020-09-22
Fix FA + triangles.

vladotomov:
Simply uses the old code for triangles + FA.


cervenal (approved):
Fixes FA on triangles for me. Thanks.

camierjs (approved):

camierjs:
LGTM

PR #94, opened 2020-09-24
Merge X and V bases

dylan-copeland:
This PR adds three new ways to merge `X-X0` and `V` bases. 

1. `-romxv` (specified in offline and online runs): use `V` basis for `X-X0`.
2. `-romvx` (specified in offline and online runs) and `-efx` to specify an energy fraction for `X-X0` only, default 0.9999 (offline only): use `X-X0` basis for `V`, usually requiring `-efx` closer to 1 than the usual `-ef` in order to ensure a sufficiently large basis for `X-X0` to be used also for `V`. This option is tricky and not recommended, since it is unclear how to ensure `V` uses a good basis.
3. `-romxandv` (specified in offline and online runs) and `-efx` to specify an energy fraction for merged `X-X0` and `V` basis vectors (online only).

Supporting the new options for basis merging introduces some undesirable coding complications. In particular, merging bases implies that dimensions and the number of samples can change in the ROM_Basis constructor. These parameters are usually constant inputs, but now they change and have to be copied back to `main()`. Also, the sample factors are now passed into the ROM_Basis constructor to recompute the number of samples from the new basis dimensions. I am open to suggestions if anyone has ideas for better implementation. One option is to let the ROM_Basis constructor modify the input parameters, passed by reference, but that seems like a bad policy for most of the parameters.

It seems unnecessary to specify -rdimx and -rdimv when using -romxandv to merge X and V bases, since the -efx energy fraction will be used to determine the basis size for X and V. However, specifying -rdimx and -rdimv does in fact make a huge difference. The reason is that not setting those parameters means that the default value of -1 will be used, in which case the entire basis for X and V will be sampled by the generator for X and V. Somehow the resulting basis is not as good as the basis computing by using only -rdimx and -rdimv basis vectors for X and V, respectively, where the dimensions specified correspond to a reasonable -ef value for X and V. In other words, the basis vectors for X and V corresponding to very small singular values should not be included when computing the SVD for the merged X and V basis vectors. Also, basis construction is generally sensitive to small changes (serial vs. parallel, number of "basis snapshots", etc.). 

The restore phase for -romxandv requires computing SVD for X and V on the fly, as the basis computed in the online phase is not written to file. This is not the ideal way for the online and restore phases to work, but I'm not sure this should be optimized unless the -romxandv option will be used regularly, which I doubt will happen for now.

Thus there are some ways in which this PR is not perfectly complete, but I'm not sure it is worth spending the time now to make further improvements considering that it is unclear how useful or commonly used these new basis options will be. Improvements can always be made later as needed.

@siuwuncheung Note that options -romxv and -romvx will not work with parametric ROM, since the X or V basis is not generated in those cases, in the offline stage. However, -romxandv should work, since it generates a new basis for X and V in the online stage. For parametric ROM, it seems that simple changes could be made to ignore the X or V basis.

The sedov-blast-merge directory was created rather than putting those tests in sedov-blast, because it takes too long when all the tests are in that one directory.

I had to make lots of commits to get the regression tests working properly. A squash merge seems best for this PR.


chldkdtn:
@dylan-copeland introducing simple regressions tests for these option will be helpful.

kevinhkhuynh:
> 
> 
> This PR adds three new ways to merge `X-X0` and `V` bases.
> 
>     1. `-romxv` (specified in offline and online runs): use `V` basis for `X-X0`.
> 
>     2. `-romvx` (specified in offline and online runs) and `-efx` to specify an energy fraction for `X-X0` only, default 0.9999 (offline only): use `X-X0` basis for `V`, usually requiring `-efx` closer to 1 than the usual `-ef` in order to ensure a sufficiently large basis for `X-X0` to be used also for `V`. This option is tricky and not recommended, since it is unclear how to ensure `V` uses a good basis.
> 
>     3. `-romxandv` (specified in offline and online runs) and `-efx` to specify an energy fraction for merged `X-X0` and `V` basis vectors (online only).
> 
> 
> Supporting the new options for basis merging introduces some undesirable coding complications. In particular, merging bases implies that dimensions and the number of samples can change in the ROM_Basis constructor. These parameters are usually constant inputs, but now they change and have to be copied back to `main()`. Also, the sample factors are now passed into the ROM_Basis constructor to recompute the number of samples from the new basis dimensions. I am open to suggestions if anyone has ideas for better implementation. One option is to let the ROM_Basis constructor modify the input parameters, passed by reference, but that seems like a bad policy for most of the parameters.
> 
> It seems unnecessary to specify -rdimx and -rdimv when using -romxandv to merge X and V bases, since the -efx energy fraction will be used to determine the basis size for X and V. However, specifying -rdimx and -rdimv does in fact make a huge difference. The reason is that not setting those parameters means that the default value of -1 will be used, in which case the entire basis for X and V will be sampled by the generator for X and V. Somehow the resulting basis is not as good as the basis computing by using only -rdimx and -rdimv basis vectors for X and V, respectively, where the dimensions specified correspond to a reasonable -ef value for X and V. In other words, the basis vectors for X and V corresponding to very small singular values should not be included when computing the SVD for the merged X and V basis vectors. Also, basis construction is generally sensitive to small changes (serial vs. parallel, number of "basis snapshots", etc.).
> 
> The restore phase for -romxandv requires computing SVD for X and V on the fly, as the basis computed in the online phase is not written to file. This is not the ideal way for the online and restore phases to work, but I'm not sure this should be optimized unless the -romxandv option will be used regularly, which I doubt will happen for now.
> 
> Thus there are some ways in which this PR is not perfectly complete, but I'm not sure it is worth spending the time now to make further improvements considering that it is unclear how useful or commonly used these new basis options will be. Improvements can always be made later as needed.
> 
> @siuwuncheung Note that options -romxv and -romvx will not work with parametric ROM, since the X or V basis is not generated in those cases, in the offline stage. However, -romxandv should work, since it generates a new basis for X and V in the online stage. For parametric ROM, it seems that simple changes could be made to ignore the X or V basis.
> 
> The sedov-blast-merge directory was created rather than putting those tests in sedov-blast, because it takes too long when all the tests are in that one directory.
> 
> I had to make lots of commits to get the regression tests working properly. A squash merge seems best for this PR.

I am fine with the new ROM_Basis constructor and explicitly copying the basis dimensions back into main(). I think it is more clear than doing changes by reference within a constructor.

chldkdtn (approved):

kevinhkhuynh (approved):

siuwuncheung (approved):

PR #99, opened 2020-10-15
rom-dev absolute test option

kevinhkhuynh:
Add absolute fom tests (Option -a). Only compares the last timestep's solution and ignores the number of time steps (as some tests differ)

The tolerance between the FOM models for many of the tests had to be increased. Same for the rel error tol which had to be increased above 1e-1 for some of the tests to get them to pass.

None of the FOM tests are able to get below a threshold of 1e-14. Most likely it's because the solution vectors are all precision 16 and the small changes add up.

Please feel free to change the tolerances for your tests.


tests/computeSpeedup.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <cassert>
+#include <iostream>
+
+using namespace std;
+
+int main(int argc, char *argv[]) {
+    double offlineSpeed = stod(argv[1]);
+    double onlineSpeed = stod(argv[2]);
+    double errorBound = stod(argv[3]);
+    if (offlineSpeed / onlineSpeed * 100 < errorBound) {
```
dylan-copeland: @kevinhkhuynh My understanding is that `offlineSpeed` is the offline run-time, and `onlineSpeed` is the online run-time. If one expects the online run to be 2.1x as fast as the offline run, then `errorBound` should be set to 210. If that understanding is correct, then I would propose some small changes, as a matter of taste. First, I would eliminate the percentage (multiplication by 100) and just set a speed-up factor, e.g. 2.1. That seems simpler to me. Also, `errorBound` is a term copied from `checkError.cpp`, and a name like `speedupMinimum` or `speedupTol` would be more descriptive. Also, `offlineSpeed` should be `offlineTime` (similarly for online).

PR #100, opened 2020-10-16
 Parametric time windowing by number of samples

siuwuncheung:
This PR implements the time windowing approach by the number of snapshots in the parametric ROM setting. 

In each single run, the options of `-romos`, `-rostype`, `-romsrhs` (=`-rhs` in merge) have to be identical in the offline, merge and online phases. Moreover, `nwin` and `-tw` have to be default-valued in both the offline and merge phases. The options `-nwinsamp` and `-nwinover` are user-defined numbers of samples and overlapping samples in the time windows in the merge phase. 

In the offline phase, we run
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -rostype previous -romsrhs -bef 1.0 -rpar 0`
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -rostype previous -romsrhs -bef 1.2 -rpar 1`
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -offline -romsvds -romos -rostype previous -romsrhs -bef 0.8 -rpar 2`

The snapshots are then merged by
`./merge -nset 3 -ef 0.9999 -rhs -nwinsamp 10 -nwinover 8 -romos -rostype previous`

A reference solution is given by the full-order model by running
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -bef 0.5 -writesol -visit`

Then in the online phase, we run
`srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -online -romhr -romos -sfacx 1 -sfacv 32 -sface 32 -soldiff -romgs -romsrhs -bef 0.5 -rparos -rostype previous -nwin 25 -twp twpTemp.csv`


siuwuncheung:
@chldkdtn @dylan-copeland @kevinhkhuynh This PR is ready for review. Thanks.

dylan-copeland:
@siuwuncheung It looks like `git pull origin rom-dev` is needed.

dylan-copeland:
@siuwuncheung Is my understanding correct, that the number of samples defining a window is not specified in the Laghos offline run, so that there are no windows in Laghos? This means that Laghos outputs snapshots (rather than a basis, since the parameter ID is specified), and then `merge` handles the construction of windows?

laghos.cpp
```diff
@@ -1293,6 +1293,7 @@ int main(int argc, char *argv[])
                     timeLoopTimer.Start();
 
                     basis->ProjectFOMtoROM(S, romS);
+                    cout << myid << ": initial romS norm " << romS.Norml2() << endl;
```
dylan-copeland: How about doing `cout` only `if (myid == 0)`? Also, it would be clearer to output the window index in this line. That is, rather than `initial romS norm` write `window " << romOptions.window << " initial romS norm "`.

merge.cpp
```diff
@@ -46,7 +69,113 @@ void LoadSampleSets(const int rank, const double energyFraction, const int nsets
     {
         cout << varName << " basis summary output: ";
         BasisGeneratorFinalSummary(basis_generator.get(), energyFraction, cutoff);
-        PrintSingularValues(rank, basename, varName, basis_generator.get(), usingWindows, window);
+        PrintSingularValues(rank, basename, varName, basis_generator.get(), usingWindows, basisWindow);
+    }
+}
+
+void MergeSamplingTimeWindow(const int rank, const double energyFraction, const int nsets, const std::string& basename, VariableName v, const std::string& varName, const std::string& basis_filename,
```
dylan-copeland: This lines are long (fewer arguments per line would be better).

merge.cpp
```diff
@@ -46,7 +69,113 @@ void LoadSampleSets(const int rank, const double energyFraction, const int nsets
...
+
+    if (rank == 0)
+    {
+        cout << varName << " basis summary output: ";
+        BasisGeneratorFinalSummary(window_basis_generator.get(), energyFraction, cutoff);
+        PrintSingularValues(rank, basename, varName, window_basis_generator.get(), true, basisWindow);
+    }
+}
+
+void LoadSampleSets(const int rank, const double energyFraction, const int nsets, const std::string& basename, VariableName v, const bool usingWindows, const int windowNumSamples,
```
dylan-copeland: Long lines

merge.cpp
```diff
@@ -46,7 +69,113 @@ void LoadSampleSets(const int rank, const double energyFraction, const int nsets
...
+        varName = "Fe";
+        break;
+    default:
+        varName = "X";
+    }
+    std::string basis_filename = basename + "/basis" + varName + std::to_string(basisWindow);
+
+    if (windowNumSamples > 0)
+    {
+        MergeSamplingTimeWindow(rank, energyFraction, nsets, basename, v, varName, basis_filename, windowOverlapSamples, basisWindow, useOffset, offsetType, dim, totalSamples, offsetAllWindows, cutoff);
```
dylan-copeland: Long line

siuwuncheung:
> @siuwuncheung Is my understanding correct, that the number of samples defining a window is not specified in the Laghos offline run, so that there are no windows in Laghos? This means that Laghos outputs snapshots (rather than a basis, since the parameter ID is specified), and then `merge` handles the construction of windows?

Yes it is correct. All the snapshots are taken in the same "sampling window" for each offline parameter, and `merge` classifies the snapshots into multiple "basis windows", and then merges the snapshots across the parameters and computes the basis in each "basis window".

merge.cpp
```diff
@@ -61,6 +190,148 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+
+    tSnap.clear();
+    double t = 0.0;
+    while (infile_tSnap >> t)
+    {
+        tSnap.push_back(t);
+    }
+}
+
+void GetParametricTimeWindows(const int nset, const bool rhsBasis, const std::string& basename, const int windowNumSamples, int &numBasisWindows, Array<double> &twep, std::vector<std::vector<int>> &offsetAllWindows)
```
dylan-copeland: Long line

laghos.cpp
siuwuncheung: @dylan-copeland I agree. Let's modify it this way.

merge.cpp
```diff
@@ -46,7 +69,116 @@ void LoadSampleSets(const int rank, const double energyFraction, const int nsets
     {
         cout << varName << " basis summary output: ";
         BasisGeneratorFinalSummary(basis_generator.get(), energyFraction, cutoff);
-        PrintSingularValues(rank, basename, varName, basis_generator.get(), usingWindows, window);
+        PrintSingularValues(rank, basename, varName, basis_generator.get(), usingWindows, basisWindow);
+    }
+}
+
```
dylan-copeland, line 75: @siuwuncheung Could you please add some documentation explaining the overall algorithm for merging parametric bases when defined by number of samples or physical time, as well as a description of what these new functions do in that algorithm? It is not very easy to see the algorithm from the code.

merge.cpp
```diff
@@ -61,6 +193,160 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+                tTemp[paramID+nset*VariableName::Fv] = tSnapFv[paramID].back();
+                tTemp[paramID+nset*VariableName::Fe] = tSnapFe[paramID].back();
+            }
+        }
+
+        double windowLeft = *max_element(tTemp.begin(), tTemp.end());
+        double overlapMidpoint = (windowLeft + windowRight) / 2;
+        twepTemp.push_back(overlapMidpoint);
+
+        if (numSnap[0] == offsetCurrentWindow[0]+1)
```
dylan-copeland, line 334: This checks only variable `X`. Should all variables be checked to see if their samples have been used up, in which case this should be the final window?

merge.cpp
```diff
@@ -61,6 +193,160 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+    }
+}
+
+void GetParametricTimeWindows(const int nset, const bool rhsBasis, const std::string& basename, const int windowNumSamples, int &numBasisWindows,
+                              Array<double> &twep, std::vector<std::vector<int>> &offsetAllWindows)
+{
+    std::vector<double> tVec;
+    std::vector<std::vector<double>> tSnapX, tSnapV, tSnapE, tSnapFv, tSnapFe;
+    const int numVar = (rhsBasis) ? 5 : 3;
+    std::vector<int> numSnap(nset*numVar);
```
dylan-copeland, line 216: It may be simpler for `numSnap` to be `std::vector<std::vector<int>>`, so that 2D indices can be used.

merge.cpp
```diff
@@ -61,6 +193,160 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+            reverse(tVec.begin(), tVec.end());
+            tSnapFe.push_back(tVec);
+            numSnap[paramID+nset*VariableName::Fe] = tVec.size();
+        }
+    }
+
+    bool lastBasisWindow = false;
+    std::vector<double> tTemp(nset*numVar, 0.0);
+    std::vector<int> offsetCurrentWindow(nset*numVar, 0);
+    std::vector<double> twepTemp;
```
dylan-copeland: Can `twepTemp` be eliminated by just pushing values directly to `twep`?

merge.cpp
```diff
@@ -61,6 +193,160 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+            GetSnapshotTime(paramID, basename, "Fe", tVec);
+            reverse(tVec.begin(), tVec.end());
+            tSnapFe.push_back(tVec);
+            numSnap[paramID+nset*VariableName::Fe] = tVec.size();
+        }
+    }
+
+    bool lastBasisWindow = false;
+    std::vector<double> tTemp(nset*numVar, 0.0);
+    std::vector<int> offsetCurrentWindow(nset*numVar, 0);
```
dylan-copeland, line 248: See comment above, to make this a 2D vector.

merge.cpp
siuwuncheung: @dylan-copeland Dylan, thanks for asking this question. At first I intended not to define `twepTemp`, but `twep` is of type `Array <double>`, which is used in `ReadTimeWindows`, and I suppose it is better not to change the data type. In `GetParametricTimeWindows`, the final size of `twep`, which is the number of basis windows, is not known in prior. The only way I can only think of is using a STL vector `std::vector<double> twepTemp` to record the elements incrementally, and then set `Array<double> twep` of the same size and make a copy of every element to `twep`. May I ask if there is a better way round?

merge.cpp
siuwuncheung: @dylan-copeland I am thinking if it is better or not to use a 2D vector for `tTemp` and `offsetCurrentWindow`. We need to find the maximum and minimum of recorded time instants. Right now we use `std::max_element` on `std::vector<double> tTemp`. I think the same does not apply to a 2D vector. On the other hand, we need to stack `offsetCurrentWindow` up to `offsetAllWindows`, which will become 3D. I think we can change the data type of `numSnap` to be a 2D vector though.

merge.cpp
dylan-copeland: We could change `ReadTimeWindows` to have `twep` as `std::vector`. There is not a good reason to use `Array`. If you want to make this change, that is fine, but I don't think it is really important. Either way is fine with me.

dylan-copeland: Another option is to use `Array::Append` (see mfem/general/array.hpp).

merge.cpp
siuwuncheung: @dylan-copeland After thinking about the whole procedure, I think it is better to keep `numSnap` and `offsetCurrentWindow` to have the same data type and structure, for easier comparison at the final time window, i.e. `numSnap[i] == offsetCurrentWindow[i] + 1`

merge.cpp
siuwuncheung: @dylan-copeland Thanks Dylan. Got it!

merge.cpp
```diff
@@ -158,48 +465,53 @@ int main(int argc, char *argv[])
...
-            LoadSampleSets(myid, energyFraction, nset, outputPath, "Fe", usingWindows, t, dimE, totalSnapshotSizeFe, cutoff[4]);
-        }
+            LoadSampleSets(myid, energyFraction, nset, outputPath, VariableName::X, usingWindows, windowNumSamples, windowOverlapSamples, basisWindow, useOffset, trueOffsetType, dimX, totalSnapshotSize, offsetAllWindows, cutoff[0]);
+            LoadSampleSets(myid, energyFraction, nset, outputPath, VariableName::V, usingWindows, windowNumSamples, windowOverlapSamples, basisWindow, useOffset, trueOffsetType, dimV, totalSnapshotSize, offsetAllWindows, cutoff[1]);
+            LoadSampleSets(myid, energyFraction, nset, outputPath, VariableName::E, usingWindows, windowNumSamples, windowOverlapSamples, basisWindow, useOffset, trueOffsetType, dimE, totalSnapshotSize, offsetAllWindows, cutoff[2]);
 
-        if (myid == 0 && usingWindows)
-        {
-            outfile_twp << twep[t] << ", ";
+            args.AddOption(&windowOverlapSamples, "-nwinover", "--numwindowoverlap", "Number of samples for ROM window overlap.");
```
dylan-copeland: Should this be here? It looks like a copy and paste mistake.

merge.cpp
```diff
@@ -61,6 +193,154 @@ void GetSnapshotDim(const int id, const std::string& basename, const std::string
...
+                        offsetCurrentWindow[paramID+nset*VariableName::Fe] += 1;
+                    }
+                }
+            }
+        }
+        offsetAllWindows.push_back(offsetCurrentWindow);
+        numBasisWindows += 1;
+
+        // Find the largest time, windowLeft, such that the last snapshot is counted for every variable and parameter
+        // The next basis window takes this snapshot and is opened at the midpoint of windowLeft and windowRight,
```
dylan-copeland: Comma at the end.

dylan-copeland:
@siuwuncheung For the example test you described for this PR, the FOM solution is produced by running `srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -bef 0.5 -writesol -visit`. The visit files are written to the main `Laghos` directory, with names like `_000179`. Should there be a prefix, and should these be in a subdirectory? Is this a change made in this PR, or was it already like this? Also, the online run using the parametric basis has error 8% for velocity. Is this error relative to the FOM simulation run with `-bef 0.5`?

siuwuncheung:
> @siuwuncheung For the example test you described for this PR, the FOM solution is produced by running `srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -bef 0.5 -writesol -visit`. The visit files are written to the main `Laghos` directory, with names like `_000179`. Should there be a prefix, and should these be in a subdirectory? Is this a change made in this PR, or was it already like this? Also, the online run using the parametric basis has error 8% for velocity. Is this error relative to the FOM simulation run with `-bef 0.5`?

@dylan-copeland I think I did not change the code for writing the solution. I also see these directories while running the code on my laptop, but I don't recall exactly when it starts, probably it had been like this for few PRs. But as far as I remember I don't see these directories in `Laghos` main directory on LC machines.

kevinhkhuynh:
> 
> 
> > @siuwuncheung For the example test you described for this PR, the FOM solution is produced by running `srun -n 1 -p pdebug laghos -m data/cube01_hex.mesh -pt 211 -tf 0.1 -bef 0.5 -writesol -visit`. The visit files are written to the main `Laghos` directory, with names like `_000179`. Should there be a prefix, and should these be in a subdirectory? Is this a change made in this PR, or was it already like this? Also, the online run using the parametric basis has error 8% for velocity. Is this error relative to the FOM simulation run with `-bef 0.5`?
> 
> @dylan-copeland I think I did not change the code for writing the solution. I also see these directories while running the code on my laptop, but I don't recall exactly when it starts, probably it had been like this for few PRs. But as far as I remember I don't see these directories in `Laghos` main directory on LC machines.

Oh, about this. This is my fault. I had fixed it last week in the other PR about absolute tests, but it isn't merged in yet.

dylan-copeland (approved):

chldkdtn:
@siuwuncheung would you add some regression tests related to this PR? Ideally, the one that takes not much time.

siuwuncheung:
> @siuwuncheung would you add some regression tests related to this PR? Ideally, the one that takes not much time.

@chldkdtn I plan to add the regression tests in PR #101, for various offset styles and for both physical time windowing and by number of snapshots. May I ask if you think it is okay or you want tests added in this PR?

kevinhkhuynh (approved):

chldkdtn:
> > @siuwuncheung would you add some regression tests related to this PR? Ideally, the one that takes not much time.
> 
> @chldkdtn I plan to add the regression tests in PR #101, for various offset styles and for both physical time windowing and by number of snapshots. May I ask if you think it is okay or you want tests added in this PR?

@siuwuncheung You do not need to introduce many regression tests. Let's add only one regression test in this PR.

siuwuncheung:
> > > @siuwuncheung would you add some regression tests related to this PR? Ideally, the one that takes not much time.
> > 
> > 
> > @chldkdtn I plan to add the regression tests in PR #101, for various offset styles and for both physical time windowing and by number of snapshots. May I ask if you think it is okay or you want tests added in this PR?
> 
> @siuwuncheung You do not need to introduce many regression tests. Let's add only one regression test in this PR.

@chldkdtn Got it. I will do it as soon as possible, and we can check that the regression tests other than the newly added ones all pass then.

siuwuncheung:
@chldkdtn The regression tests are added and PR #101 is merged into this PR as it just involves changes in a few places. Please note that the regression tests on Sedov Blast in the baseline branch fail because of some changes in command lines inputs, e.g. removal of `rostype` in `laghos.cpp` and the new tags of `romos` and `rostype` in merge. By manually running the same tests and comparing the results with the confluence page, the results are not significantly changed.

chldkdtn:
LGTM

chldkdtn (approved):
@siuwuncheung Great job, Tony! Make sure LGTM passes all the tests before you merge this PR.

dylan-copeland:
@siuwuncheung The merging of this PR into rom-dev seems to have introduced regression tests with `numsamp` in their name, which fail on `merge.cpp:395`: 
`Verification failed: (windowNumSamples == 0 || numWindows == 0) is false:`

siuwuncheung:
@dylan-copeland Thanks for pointing that out. I did a regression test on the rom-dev branch after merge and find the same. This is a new regression test introduced in this PR. I will quickly fix it with a small PR.

PR #107, opened 2020-10-24
Added internal energy calculation for partial assembly

varchasgopalaswamy:
Hi,

I noticed that the total internal energy calculation was only enabled for full assembly. I thought it might be (and found it) useful to have it in PA as well, so I made the following edits. If it looks good I can also add a similar one I have for kinetic energy (and therefore total energy), so the user could see how well energy is being conserved. 

- Varchas

### DIFF
--- a/laghos_solver.cpp
+++ b/laghos_solver.cpp
@@ -563,25 +563,37 @@ void LagrangianHydroOperator::ComputeDensity(ParGridFunction &rho) const
    }
 }
 
+
 double LagrangianHydroOperator::InternalEnergy(const ParGridFunction &gf) const
 {
    double glob_ie = 0.0;
-   // This should be turned into a kernel so that it could be displayed in pa
-   if (!p_assembly)
-   {
-      Vector one(l2dofs_cnt), loc_e(l2dofs_cnt);
-      one = 1.0;
-      Array<int> l2dofs;
-      double loc_ie = 0.0;
-      for (int e = 0; e < NE; e++)
-      {
-         L2.GetElementDofs(e, l2dofs);
-         gf.GetSubVector(l2dofs, loc_e);
-         loc_ie += Me(e).InnerProduct(loc_e, one);
-      }
-      MPI_Comm comm = H1.GetParMesh()->GetComm();
-      MPI_Allreduce(&loc_ie, &glob_ie, 1, MPI_DOUBLE, MPI_SUM, comm);
-   }
+   
+    // get the restriction and interpolator objects 
+    const QuadratureInterpolator* l2_interpolator = L2.GetQuadratureInterpolator(ir);
+    l2_interpolator->SetOutputLayout(QVectorLayout::byVDIM);
+    auto L2r = L2.GetElementRestriction(ElementDofOrdering::LEXICOGRAPHIC);
+    const int NQ = ir.GetNPoints(); 
+   Vector e_vector(NE*NQ), eintQ(NE*NQ);
+
+    // Get internal energy at the quadrature points
+    L2r->Mult(gf, e_vector);
+    l2_interpolator->Values(e_vector, eintQ);
+
+    // Get the IE, initial weighted mass 
+    auto ie_vals = mfem::Reshape(eintQ.Read(),NQ, NE);
+    auto initial_mass = mfem::Reshape(qdata.rho0DetJ0w.Read(),NQ, NE);
+
+
+    double internal_energy = 0;
+    for (int e = 0; e < NE; ++e) {
+        for (int q = 0; q < NQ; ++q) {
+            
+            internal_energy += initial_mass(q,e)*ie_vals(q,e);
+
+        }
+    }
+   MPI_Allreduce(&internal_energy, &glob_ie, 1, MPI_DOUBLE, MPI_SUM, L2.GetParMesh()->GetComm());
+
    return glob_ie;
 }
 
--- a/laghos_solver.cpp
+++ b/laghos_solver.cpp
@@ -635,7 +635,8 @@ double LagrangianHydroOperator::InternalEnergy(const ParGridFunction &gf) const
     l2_interpolator->SetOutputLayout(QVectorLayout::byVDIM);
     auto L2r = L2.GetElementRestriction(ElementDofOrdering::LEXICOGRAPHIC);
     const int NQ = ir.GetNPoints(); 
-   Vector e_vector(NE*NQ), eintQ(NE*NQ);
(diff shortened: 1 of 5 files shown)


vladotomov:
Thank you Varchas, very nice work. Could you also do it for the KE and then we can merge it.

laghos_solver.cpp
```diff
@@ -563,25 +563,37 @@ void LagrangianHydroOperator::ComputeDensity(ParGridFunction &rho) const
...
-   }
+   
+    // get the restriction and interpolator objects 
+    const QuadratureInterpolator* l2_interpolator = L2.GetQuadratureInterpolator(ir);
+    l2_interpolator->SetOutputLayout(QVectorLayout::byVDIM);
+    auto L2r = L2.GetElementRestriction(ElementDofOrdering::LEXICOGRAPHIC);
+    const int NQ = ir.GetNPoints(); 
+   Vector e_vector(NE*NQ), eintQ(NE*NQ);
+
+    // Get internal energy at the quadrature points
```
camierjs, line 641: Thanks for the addition @varchasgopalaswamy.
Once the `cpu` code is done, we should turn the for-loops into kernels that could run on the `mfem::Device`s.
Let me know if you need extra information for that.

varchasgopalaswamy: So I took a look at how the MFEM_FORALL loops work, and I think I have it working - except that I am not sure how to do reductions inside an MFEM_FORALL. So if I have something like 

```
double integral = 0; 
MFEM_FORALL_2D(e, NE, Q1D, Q1D, 1,
      {
         MFEM_FOREACH_THREAD(qy,y,Q1D)
         {
            MFEM_FOREACH_THREAD(qx,x,Q1D)
            {
               const int q = qx + qy * Q1D;
               double vmag = 0; 
               for(int k = 0; k < VDIM; k++){
                  vmag += pow(f_vals(k,q,e),ln_norm);
               }
               integral += mass_vals(q,e)*vmag;
            }
         }
      });

```

I get the compiler error 
```laghos_solver.cpp:616:25: error: assignment of read-only variable ‘integral’
         integral += mass_vals(q,e)*vmag;
         ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

which I suppose makes sense since `integral` is defined outside the kernel. Are there any flags I should pass when instantiating `double integral = 0;` or when creating the `MFEM_FORALL` loop?

camierjs: You can look at what is done in [Rho0DetJ0Vol](https://github.com/CEED/Laghos/blob/16fa28294908fd15dda8241d6d6c75ba50a12b50/laghos_solver.cpp#L83) to see how the reduction can be done.

Before entering the `MFEM_FORALL` you must also take care of the pointers by using the appropriate `Read`/`ReadWrite` and `Reshape` if needed: this will help with the compiler error you show.

varchasgopalaswamy: Oh, I think I see. The reduction needs to be done like `volume = vol * one;`?

camierjs: Right - it's a possible way to do, not the only one.

varchasgopalaswamy:
Alright, I've pushed some changes that wrap the integral in an `MFEM_FORALL` loop and add the corresponding kinetic energy calculation, and it seems to be working as expected.

camierjs:
The kernel looks good! Have you tried it with some devices (`debug`, `cuda`)?
The output might be too long now with these new outputs: what do you think @vladotomov ?
We also need to check the Travis tests.

vladotomov:
Thank you Varchas! I'd prefer to keep the old output. Could you please restore that?

vladotomov:
The 1D run
`mpirun -np 6 ./laghos -p 2 -dim 1 -rs 5 -tf 0.2 -fa`
doesn't work for me anymore.

`MFEM abort: Unknown kernel
 ... in function: void mfem::D2QValues(const mfem::FiniteElementSpace &, const mfem::DofToQuad *, const mfem::Vector &, mfem::Vector &)
 ... in file: fem/quadinterpolator.cpp:679`

varchasgopalaswamy:
Hi Vladimir,
Sorry for the delay in getting back to you.

1) Yes, I'll restore the old output. 
2)  I did not test my changes in 1D (my bad...), so I didn't catch this. It looks like the problem is when I interpolate from the DOFs to the quadrature points in `laghos_solver.cpp:642`. The error comes up in `fem/quadinterpolator.cpp:679`, where it looks like `QuadratureInterpolator::Values` is only implemented for 2D and 3D meshes. 

Is there an alternate way to perform the interpolation when the mesh is 1D? I looked through the MFEM issues list but it didn't seem like anyone else had run into this. 

- Varchas

vladotomov:
Thank you Varchas.

The 1D version of `Values` is something that we need to add to mfem, don't worry about it.

laghos_solver.cpp
```diff
@@ -563,39 +563,113 @@ void LagrangianHydroOperator::ComputeDensity(ParGridFunction &rho) const
...
-   }
-   return glob_ke;
+  double glob_ke = 0.0;
+   
+    // get the restriction and interpolator objects 
+    const QuadratureInterpolator* h1_interpolator = H1.GetQuadratureInterpolator(ir);
+    h1_interpolator->SetOutputLayout(QVectorLayout::byVDIM);
+    auto H1r = H1.GetElementRestriction(ElementDofOrdering::LEXICOGRAPHIC);
+    const int NQ = ir.GetNPoints(); 
+   Vector e_vector(dim*NE*NQ), ekinQ(dim*NE*NQ);
```
vladotomov: Shouldn't the E-vector be over dofs, not quad points? I'm guessing that in the next two calls the `Mult` writes at the correct indices, and the `Values` reads the same indices.

varchasgopalaswamy: You're saying that I should have written `e_vector(dim*NE*NQ)` as `e_vector(dim*NE)`? And that the next few lines will read/write from the first dim*NE block, so it didn't trigger a segfault or anything like that?

vladotomov:
The mfem PR for the 1D error is [here](https://github.com/mfem/mfem/pull/1921).

laghos_solver.cpp
vladotomov: I think it should be
```
const int ND = H1.GetFE(0)->GetDof();
Vector e_vector(dim*NE*ND);
```
And similar change in the internal energy function.

vladotomov (approved):

vladotomov:
@camierjs this seems ready to go in, please take one more look.

camierjs (approved):

PR #109, opened 2020-10-29
change rdimfv in sedov-blast-merge tests

kevinhkhuynh:
Changing GNAT.C asserts to verify causes an error as rdimfv is greater than the number of columns in the basis matrix. So the numbers have been changed accordingly to fit the size of the basis matrix.

### DIFF
--- a/tests/sedov-blast-merge/sedov-blast-mergexv.sh
+++ b/tests/sedov-blast-merge/sedov-blast-mergexv.sh
@@ -5,7 +5,7 @@ case $subTestNum in
     $LAGHOS -m data/cube01_hex.mesh -pt 211 -tf 0.05 -offline -ef 0.9999 -writesol -romsvds -romos -rostype load -romsrhs -romxandv
     ;;
   2)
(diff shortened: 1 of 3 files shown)


dylan-copeland (approved):
Thanks @kevinhkhuynh , I ran into this issue in the past but did not fix it. It's nice to have it fixed now.

siuwuncheung (approved):

chldkdtn:
@kevinhkhuynh I am wondering why we had rdimfv greater than the number of columns in the basis matrix in the beginning. Can you explain?

kevinhkhuynh:
> 
> 
> @kevinhkhuynh I am wondering why we had rdimfv greater than the number of columns in the basis matrix in the beginning. Can you explain?

Before, if rdimfv was greater than the column in the basis matrix, libROM would not throw an error, but now it does, so before it was doing something undefined in GNAT. I believe the rdimfv number was just a typo, or something changed in later development that caused rdimfv to decrease slightly.

@dylan-copeland Is that correct?

dylan-copeland:
@kevinhkhuynh @chldkdtn I think we had a Sedov test with large error when the basis dimensions recommended by the basis generation summary (offline) were used, so we doubled the basis dimensions for better accuracy. For Fv, this turned out to be too large, by accident.

chldkdtn (approved):

PR #114, opened 2020-11-06
Add Rayleigh-Taylor Instability Problem to rom branch

siuwuncheung:

Velocity at t = 4.0
velocity

Energy at t = 4.0
energy

Density at t = 4.0
density


rom/laghos_solver.cpp
```diff
@@ -833,6 +832,434 @@ void LagrangianHydroOperator::UpdateQuadratureData(const Vector &S) const
     timer.quad_tstep += nzones;
 }
 
+/// Trace of a square matrix
+template<int H, int W, typename T>
+//MFEM_HOST_DEVICE inline
```
dylan-copeland: Why does this need to be commented out? It seems to work on Mac.

dylan-copeland:
Are the added mesh files rt2D.mesh, square01_quad_unstr.mesh, square01_tri.mesh necessary? I don't see where they are used.

dylan-copeland:
@siuwuncheung The regression tests all seem to be failing, and I suppose it is because of small differences (e.g. number of timesteps) due to round-off level changes caused by updating the FOM code from the master branch of Laghos. If so, that is fine, but can you please report whether you have verified this is the case? It is a lot of work to check every regression test output file to verify that the change is acceptable, so I think it is fine if just one person (usually the PR author) does this and reports the results.

chldkdtn:
@siuwuncheung please see if you get a seg fault when running the following simulation:

./laghos -p 7 -m data/rt2D.mesh -tf 4 -rs 1 -ok 4 -ot 3 -offline -romsvds -ef 0.9999 -nwinsamp 5 -romos -rostype load -romsrhs -writesol -visit

I suspect that it is something to do with the end or beginning of time window exactly coincides with the end of time steps.

rom/laghos_solver.cpp
siuwuncheung: The lines were commented in attempt to find a compilation error. The error had been located somewhere else. These lines are uncommented now.

siuwuncheung:
> Are the added mesh files rt2D.mesh, square01_quad_unstr.mesh, square01_tri.mesh necessary? I don't see where they are used.

The mesh file rt2D.mesh is used for the 2D Rayleigh-Taylor instability problem. The mesh files square01_quad_unstr.mesh and square01_tri.mesh are not necessary and can be removed.

siuwuncheung:
> @siuwuncheung please see if you get a seg fault when running the following simulation:
> 
> ./laghos -p 7 -m data/rt2D.mesh -tf 4 -rs 1 -ok 4 -ot 3 -offline -romsvds -ef 0.9999 -nwinsamp 5 -romos -rostype load -romsrhs -writesol -visit
> 
> I suspect that it is something to do with the end or beginning of time window exactly coincides with the end of time steps.

@chldkdtn This happens when the offset option is on and the last time step is the only snapshot sample in the last sampling window, i.e. a sampling window is closed in the second-last time step. This leaves no snapshot vectors in the basis generator for position, velocity and energy. I think this issue is not introduced in this PR and it exists for a while but we don't have the coincidence to see this seg fault. I will fix this in this PR.

siuwuncheung:
> @siuwuncheung The regression tests all seem to be failing, and I suppose it is because of small differences (e.g. number of timesteps) due to round-off level changes caused by updating the FOM code from the master branch of Laghos. If so, that is fine, but can you please report whether you have verified this is the case? It is a lot of work to check every regression test output file to verify that the change is acceptable, so I think it is fine if just one person (usually the PR author) does this and reports the results.

@dylan-copeland Thanks for pointing that out. I will have a careful check on each regression test output.

rom/laghos.cpp
```diff
@@ -1253,15 +1252,14 @@ int main(int argc, char *argv[])
                             else
                                 outfile_twp << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2] << "\n";
                         }
-                        delete sampler;
+                        sampler = NULL;
```
dylan-copeland, line 1454: Replacing `delete sampler` with `sampler = NULL` looks like a memory leak (`sampler` will never be deleted). Should there be `delete sampler` and `sampler = NULL`?

rom/laghos.cpp
```diff
@@ -1230,7 +1230,6 @@ int main(int argc, char *argv[])
                             else
                                 outfile_twp << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2] << "\n";
                         }
-                        delete samplerLast;
```
dylan-copeland: Why should this `delete` be deleted?

chldkdtn:
@siuwuncheung we need regression tests for ROM, RT instability problems. :)

rom/laghos.cpp
siuwuncheung: Yes, `delete sampler` should be kept.

rom/laghos.cpp
siuwuncheung: Yes, `delete samplerLast` should be kept.

chldkdtn:
@siuwuncheung have you get a chance of looking into regression test failures?

PR #118, opened 2020-11-12
Add Rayleigh-Taylor Instability Problem

siuwuncheung:

Velocity at t = 4.0
velocity

Energy at t = 4.0
energy

Density at t = 4.0
density


vladotomov:
Could you add the sample run to `README`?

siuwuncheung:
> Could you add the sample run to `README`?

@vladotomov Thanks for reviewing this PR. I can add the sample run to `README`. However, I am not able to reproduce the results from the paper with the command line above so far. I am not sure if I misunderstood the problem setting. The amplitude of the "sliding" surface is much smaller than that in the paper, and there is a distortion of the mesh around the bottom-left corner and the top-right corner which does not appear in the paper. I will soon put a figure of the density at the final time in this PR. I am thinking that if it is because of the boundary condition. In this problem, it seems that we should enforce symmetry planes and impose v.n = 0 only for the vertical walls, unlike the other tests we do it on the whole boundary. If that is the case, it would take a while for me to figure out how to implement the boundary condition.

laghos_solver.cpp
```diff
@@ -782,10 +732,23 @@ void LagrangianHydroOperator::UpdateQuadratureData(const Vector &S) const
                // direction of maximal compression. This is used to define the
                // relative change of the initial length scale.
                v.GetVectorGradient(*T, sgrad_v);
+
+               double vorticity_coeff = 1.0;
+               if (use_vorticity)
+               {
+                  const double grad_norm = sgrad_v.FNorm();
+                  const double div_v = fabs(sgrad_v.Trace());
+                  vorticity_coeff = (grad_norm > 0.0) ? div_v / grad_norm : 1.0;
```
vladotomov, line 746: @camierjs could you propagate these changes (`LagrangianHydroOperator::UpdateQuadratureData`) to the quadrature kernel?

camierjs: `done`

vladotomov: Thank you!

camierjs:
There is still an issue with the `CUDA` run: let me look at it.

camierjs:
Energy at t = 3.88
Screen Shot 2020-11-30 at 4 31 39 PM

vladotomov (approved):

vladotomov:
@camierjs @tzanio are we merging?

tzanio (approved):

camierjs (approved):

PR #119, opened 2020-11-24
Update ROM to work with refactored libROM

kevinhkhuynh:
Update ROM to work with refactored libROM in coordiation with PR #50 in libROM


kevinhkhuynh:
@dylan-copeland The option to use a different mfem/libROM has been added. To use the functionality, run runRegressionTests.sh -l "PATH" and pass in a PATH that contains both mfem and libROM as subdirectories. If you would like to only change one of the libraries and keep the other the same (ex. test libROM but use same mfem for both baseline and target), symbolic links to the dependencies directories can be used for the library that should be the same between the baseline and target branch.

dylan-copeland (approved):

siuwuncheung (approved):

chldkdtn (approved):

PR #122, opened 2021-01-01
Remove multiple final outputs in mpirun

siuwuncheung:


rom/laghos_rom.cpp
```diff
@@ -2236,28 +2236,32 @@ void PrintNormsOfParGridFunctions(NormType normtype, const int rank, const std::
...
     double FOMnorm = sqrt(fomglob2);
     double ROMnorm = sqrt(romglob2);
     double DIFFnorm = sqrt(diffglob2);
     double relDIFFnorm = sqrt(diffglob2)/sqrt(fomglob2);
 
-    cout << rank << ": " << name << " FOM norm " << FOMnorm << endl;
-    cout << rank << ": " << name << " ROM norm " << ROMnorm << endl;
-    cout << rank << ": " << name << " DIFF norm " << DIFFnorm << endl;
-    cout << rank << ": " << name << " Rel. DIFF norm " << relDIFFnorm << endl;
+    if (rank == 0)
```
dylan-copeland, line 2245: The changes in this PR look fine, and it makes sense to print out only for rank 0, but this function seems wrong (before this PR). There is a TODO comment about some issues with this function, which we should resolve. It is not clear whether the MPI_Allreduce should be necessary in this function. It should be done in the `ParGridFunction` class. In any case, the L1 and max norms are clearly wrong (we typically use just L2).

kevinhkhuynh (approved):
I see Dylan made some comments. Just looking at the changes you've made, looks good to me.

rom/laghos_rom.cpp
chldkdtn: @dylan-copeland I will add it in "to-do-list" of our confluence page to look into this in the future (near future. :) )

chldkdtn (approved):

dylan-copeland (approved):
As noted above, this PR is fine, but we should keep in mind that the parallel norm issue should be investigated.

PR #128, opened 2021-01-25
CI: add Github Actions to replace Travis

adrienbernede:
This PR adds Github Actions, based on the work being done in MFEM (https://github.com/mfem/mfem/pull/1991)

Good:
- caching of Hypre and Metis based on archive name.
- caching of MFEM based on commit ID.

Less good:
- actions are limited in resource. Attempt to run more than 2 MPI process results in explosion of execution time.

Questions:
- [x] Travis was set up to run only on master. This is set to run on push, at least for the sake of testing this PR. Should it then be set again to run for master only?

### DIFF
--- a/.github/workflows/build-laghos.yml
+++ b/.github/workflows/build-laghos.yml
@@ -101,12 +101,6 @@ jobs:
     - name: checks 2
       run: |
         cd laghos && make checks ranks=2
-    - name: 1
-      run: |
-        cd laghos && make 1
-    - name: 2
-      run: |
-        cd laghos && make 2
     - name: tests
       run: |
         cd laghos && make tests
--- a/.github/workflows/build-laghos.yml
+++ b/.github/workflows/build-laghos.yml
@@ -101,6 +101,6 @@ jobs:
     - name: checks 2
       run: |
         cd laghos && make checks ranks=2
-#    - name: tests
-#      run: |
-#        cd laghos && make tests
+    - name: tests
+      run: |
+        cd laghos && make tests
--- a/.github/workflows/build-laghos.yml
+++ b/.github/workflows/build-laghos.yml
@@ -9,6 +9,7 @@ env:
   METIS_URL: http://glaros.dtc.umn.edu/gkhome/fetch/sw/metis/OLD
   METIS_ARCHIVE: metis-4.0.3.tar.gz
   METIS_TOP_DIR: metis-4.0.3
+  MFEM_URL: https://github.com/mfem/mfem.git
   MFEM_BRANCH: master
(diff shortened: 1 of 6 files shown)


adrienbernede:
It passed!

.github/workflows/build-laghos.yml
```diff
@@ -0,0 +1,106 @@
...
+      uses: ./laghos/.github/actions/install-mfem
+      with:
+        mfem-repo: ${{ env.MFEM_REPO }}
+        mfem-branch: ${{ env.MFEM_BRANCH }}
+        mfem-dir: ${{ env.MFEM_TOP_DIR }}
+
+    - name: make
+      run: |
+        cd laghos && make -j
+    - name: checks 1
```
camierjs: Would it be possible to split the build/checks/tests in different actions?
If not , could we rename this one from `build_laghos` to `tests_laghos` or just `tests`?

camierjs (commented):
Great work! Thanks @adrienbernede !

.github/workflows/build-laghos.yml
adrienbernede: I'm trying to understand exactly what it is you would like.
Do you mean that in the checks view you would like to have separate logs for each?

In terms of wording, 
- an "action" is just an encapsulated... action, that can be re-used elsewhere.
- in the check view, _build_ is a "job" with multiples "steps" in it, it should in fact be _build_and_test_
- _build_laghos_ is the "workflow" name. It should certainly be renamed _build_and_test_laghos_

In the end, I could try putting checks and tests in a different job. However I think that would require passing the installs as artifact, which may add complexity.

I can give it a try, let me know if that's what you meant.

camierjs: `build_and_test_laghos` would be perfect.

camierjs (approved):

camierjs:
Tests on 'PR' pushs would be useful.

adrienbernede:
Testing only on PRs targeting master.

tzanio:
Just to clarify -- we want this to run on pushes to any PR targeting master. Is that the case now?

tzanio (approved):
Awesome work, thank you so much for the quick response @adrienbernede !

camierjs:
Could the logo in the top `README` be refreshed with the status too?

adrienbernede:
@camierjs Should work. I did not remove the Travis badge yet (travis CI was not removed).
@tzanio Yes, each push to a PR targeting master.

camierjs:
Awesome! LGTM!
Thanks again!

vladotomov (approved):

PR #129, opened 2021-01-25
Maximum ROM dimension in time windowing

siuwuncheung:
A maximum ROM dimension can be set by specifying `-rdimx`, `-rdimv`, `-rdime`, `-rdimFv` and `-rdimFe` in online command line options for the time windowing cases. The true ROM dimension is taken to be the minimum over the specified number and the number of columns in the basis matrix from `ReadTimeWindowParameters`. 

Usually, in the online phase, in the serial ROM we specify the reduced dimension, and in the time windowing case we let the reduced dimensions be defined from reading the time window parameter file. Now, in this branch, if we specify the reduced dimensions in the time windowing case, it is used to control the actual ROM dimension in each window by setting the variables `romOptions.max_dim*`, and the actual dimension `romOptions.dim*` is taken as the smaller one among `romOptions.max_dim*` and the value read from the time parameter file.

This allows fair comparison of the error over the varying number of snapshot samples `-nwinsamp` by specifying the same ROM dimension and maintaining a similar speedup.


rom/laghos_utils.cpp
```diff
@@ -195,13 +195,13 @@ void split_line(const std::string &line, std::vector<std::string> &words)
 void SetWindowParameters(Array2D<int> const& twparam, ROM_Options & romOptions)
 {
     const int w = romOptions.window;
-    romOptions.dimX = twparam(w,0);
-    romOptions.dimV = twparam(w,1);
-    romOptions.dimE = twparam(w,2);
+    romOptions.dimX = min(romOptions.dimX, twparam(w,0));
```
dylan-copeland: Since `romOptions` is passed by reference, this will change `dimX` for each window, which can reduce the maximum applied to subsequent windows. You may want to have a different variable to specify the maximum.

siuwuncheung: @dylan-copeland Thanks for pointing that out. I will fix it.

siuwuncheung:
@dylan-copeland @chldkdtn @kevinhkhuynh This small PR is ready for review. Regression tests all passed.

rom/laghos.cpp
```diff
@@ -359,6 +359,11 @@ int main(int argc, char *argv[])
     const bool usingWindows = (numWindows > 0 || windowNumSamples > 0);
     if (usingWindows)
     {
+        if (romOptions.dimX  > 0) romOptions.max_dimX  = romOptions.dimX;
+        if (romOptions.dimX  > 0) romOptions.max_dimV  = romOptions.dimV;
```
dylan-copeland: Should `dimX` be `dimV`?

siuwuncheung: Yes, it was fixed but I forgot to push.

rom/laghos.cpp
```diff
@@ -359,6 +359,11 @@ int main(int argc, char *argv[])
     const bool usingWindows = (numWindows > 0 || windowNumSamples > 0);
     if (usingWindows)
     {
+        if (romOptions.dimX  > 0) romOptions.max_dimX  = romOptions.dimX;
+        if (romOptions.dimV  > 0) romOptions.max_dimV  = romOptions.dimV;
+        if (romOptions.dimE  > 0) romOptions.max_dimE  = romOptions.dimE;
+        if (romOptions.dimFv > 0) romOptions.max_dimFv = romOptions.dimFv;
+        if (romOptions.dimFe > 0) romOptions.max_dimFe = romOptions.dimFe;
```
dylan-copeland, line 371: If I understand correctly, this means that in the window case, in which we usually do not specify these dimensions, now the dimensions mean maximum dimensions. It would be helpful to add a comment or put this in the description in the corresponding `args.AddOption` calls.

siuwuncheung: Yes. Usually, in the online phase, in the serial ROM we specify the reduced dimension, and in the time windowing case we let the reduced dimensions be defined from reading the time window parameter file. Now, in this branch, if we specify the reduced dimensions in the time windowing case, it is used to control the actual ROM dimension in each window by setting the variables `romOptions.max_dim*`, and the actual dimension `romOptions.dim*` is taken as the smaller one among `romOptions.max_dim*` and the value read from the time parameter file.

dylan-copeland (approved):
This looks good to me. I did not run the regression tests and will rely on others to verify they pass.

kevinhkhuynh (approved):

chldkdtn (approved):

PR #136, opened 2021-03-04
Rom greedy algorithm

kevinhkhuynh:
Greedy algorithm implementation. Only works without hyperreduction and varies blast_energyfactor (though more parameters can be varied if we would like. I would have to add it to the PR.).

Not much has changed in laghos.cpp. Wrapped most of the code inside a do_while loop so that multiple offline/online runs could be done in succession using the greedy algorithm without affecting any of the old implementation which goes through the while loop once.

General workflow:
1. Use greedy algorithm (-build-database) to build a ROM database by passing in a set of blast_energyfactor parameter points as a file. To calculate the residual, hyperreduction can not be used.
2. After the ROM database is built, run rom-online and rom-restore using (-use-database) and passing in a new value for -bef into the command-line. Hyperreduction can be used here. The basis of the nearest ROM that was constructed in part 1 will automatically be used to run the online and restore phases.

All tests pass except for the sedov-blast-greedy test. It doesn't pass because the implementation doesn't exist on the rom branch. It can be tested by using the -t option.

This was a little messy to implement. It would be nice if we could refactor Laghos and separate the offline, online, and restore into separate functions and files. I'd be happy to work on that but it would diverge the code more from the main branch.


kevinhkhuynh:
For sedov-blast, using the ROM to compute the residual is just as slow as FOM, so I don't think the greedy algorithm will be useful for this problem.

chldkdtn:
> For sedov-blast, using the ROM to compute the residual is just as slow as FOM, so I don't think the greedy algorithm will be useful for this problem.

@kevinhkhuynh, The reason that ROM is as slow as FOM for sedov-blast is because ROM is not using hyper-reduction?

kevinhkhuynh:
Ready for review. sedov-blast-greedy.sh contains the command line for how to run the greedy algorithm.

rom/laghos_rom.cpp
```diff
@@ -192,33 +192,48 @@ void ROM_Sampler::Finalize(const double t, const double dt, Vector const& S, Arr
         }
     }
 
+    if (!writeSnapshots)
+    {
+        if (!useXV)
+        {
+            input.dimX = generator_X->getSpatialBasis()->numColumns();
```
dylan-copeland: Why is this necessary?

rom/laghos_rom.hpp
```diff
@@ -56,6 +56,15 @@ struct ROM_Options
 
     std::string *basename = NULL;
 
+    std::string basisIdentifier = "";
+    double greedyTol = 1.0;
+    double greedySat = 1.0;
+    double greedyParamSpaceMin = 0;
+    double greedyParamSpaceMax = 0;
+    int greedyParamSpaceSize = 0;
```
dylan-copeland: Isn't this a `double` in libROM?

rom/laghos_rom.hpp
```diff
@@ -56,6 +56,15 @@ struct ROM_Options
 
     std::string *basename = NULL;
 
+    std::string basisIdentifier = "";
+    double greedyTol = 1.0;
```
dylan-copeland: The greedy parameters could use some documentation. There is some in the options parsing in laghos.cpp but some more detailed descriptions could be added here.

rom/laghos.cpp
```diff
@@ -390,561 +396,657 @@ int main(int argc, char *argv[])
...
-            {
-                cout << "Laghos does not support PA in 1D. Switching to FA." << endl;
-            }
+            romOptions.basisIdentifier = "_" + to_string(paramPoints[nextSampleParameterPoint]);
+            romOptions.blast_energyFactor = paramPoints[nextSampleParameterPoint];
+        }
+    }
+    if (rom_use_database)
+    {
+        MFEM_VERIFY(!rom_build_database, "-build-database should be off when -use-database is turned on");
```
dylan-copeland, line 390: Also check for offline, online, restore?

rom/laghos_rom.cpp
kevinhkhuynh: It took me a while to remember, but I remember now. Normally, the user gives dimX, dimV, and dimE for the online phase. To calculate the residual without the user supplying this information, we need to figure it out from the offline phase, which is why this is required. This updates the romOptions dimensions for the residual.

rom/laghos.cpp
```diff
@@ -954,839 +1056,888 @@ int main(int argc, char *argv[])
...
-    if (S_old != nullptr) delete S_old;
-    if (x_gf != nullptr) delete x_gf;
-    if (v_gf != nullptr) delete v_gf;
-    if (e_gf != nullptr) delete e_gf;
-    if (rho_gf != nullptr) delete rho_gf;
-    if (vis_rho != nullptr) delete vis_rho;
-    if (vis_v != nullptr) delete vis_v;
-    if (vis_e != nullptr) delete vis_e;
-    if (visit_dc != nullptr) delete visit_dc;
-    if (v_coeff != nullptr) delete v_coeff;
```
dylan-copeland: It seems some minor changes to this file resulted in minor formatting changes to most of the lines, making it almost impossible to review. @kevinhkhuynh could you please summarize what real changes were made in this file?

rom/laghos_rom.hpp
kevinhkhuynh: Converted to int in libROM

rom/laghos_rom.cpp
kevinhkhuynh: No longer changes the input, but copies the dim to a file since the greedy algorithm is now split across multiple runs of laghos rather than doing the do-while loop.

rom/laghos.cpp
kevinhkhuynh: Removing the do while loop has reverted the formatting changes.

rom/laghos.cpp
```diff
@@ -1292,27 +1405,41 @@ int main(int argc, char *argv[])
...
+                    outfile_romS.close();
                 }
-                outfile_romS.close();
 
                 if (!romOptions.hyperreduce)
-                    basis[romOptions.window]->LiftROMtoFOM(romS, *S);
+                {
+                    if (!rom_build_database || last_step)
+                    {
+                        basis[romOptions.window]->LiftROMtoFOM(romS, *S);
```
kevinhkhuynh, line 1403: Here, we only lift every step if we're not using rom_build_database. If we are using the greedy algorithm, then only lift at the last step (which is the 2nd to last).

rom/laghos.cpp
```diff
@@ -1292,27 +1405,41 @@ int main(int argc, char *argv[])
...
+                    }
+                    if (rom_build_database && last_step)
+                    {
+                        lastLiftedSolution = *S;
+                        ode_solver_dat->Init(*oper);
+                        ode_solver_dat->Step(lastLiftedSolution, t, dt);
```
kevinhkhuynh, line 1412: At last step, initialize the ode_solver_dat and take a single step in the FOM space. This will be used for the residual.

rom/laghos.cpp
```diff
@@ -1693,8 +1823,18 @@ int main(int argc, char *argv[])
...
+    double residual;
+
     if (rom_online)
     {
+        if (rom_build_database)
+        {
+            basis[romOptions.window]->LiftROMtoFOM(romS, *S);
+            Vector residualVec = Vector(lastLiftedSolution.Size());
+            subtract(lastLiftedSolution, *S, residualVec);
+            residual = residualVec.Norml2();
```
kevinhkhuynh: Calculate the residual

rom/laghos.cpp
```diff
@@ -390,6 +406,98 @@ int main(int argc, char *argv[])
 
     NormType normtype = localmap[normtype_char];
 
+    std::vector<double> paramPoints;
+
+    CAROM::GreedyParameterPointSelector* parameterPointGreedySelector = NULL;
+
+    // If using the greedy algorithm, initialize the parameter point greedy selector.
+    if (rom_build_database)
+    {
```
dylan-copeland, line 366: This could be a function, to simplify `main()`. Also the `rom_use_database` case.

rom/laghos.cpp
```diff
@@ -990,6 +1105,7 @@ int main(int argc, char *argv[])
     // defines the Mult() method that is used by the time integrators.
     if (!rom_online) ode_solver->Init(*oper);
     if (fom_data) oper->ResetTimeStepEstimate();
+    if (rom_build_database) cout << "Blast energy factor: " << romOptions.blast_energyFactor << endl;
```
dylan-copeland: This assumes the database parameter is the energy factor, right? If this cannot be generalized, maybe it is better to remove it.

rom/tests/sedov-blast/sedov-blast-greedy.sh
```diff
@@ -0,0 +1,13 @@
+NUM_PARALLEL_PROCESSORS=8
+testNames=(build_database use_database_online use_database_restore)
+case $subTestNum in
+  1)
+    $GREEDY $LAGHOS -m data/cube01_hex.mesh -pt 211 -tf 0.01 -build-database -writesol -romsvds -greedy-param-min 0.5 -greedy-param-max 2.5 -greedy-param-size 5 -greedysubsize 2 -greedyconvsize 3
```
chldkdtn: I guess, by default, the blast energy is used as a greedy parameter. However, if we generalize this, is a command line option for which parameter will be used introduced in this PR?

chldkdtn: Also, does greedy process generate some log files recording which parameter is being used in the greedy study and which parameter value is being and has been running before and now, etc. If not, generating such a log file might be pretty helpful.

rom/laghos.cpp
```diff
@@ -1761,6 +1839,32 @@ int main(int argc, char *argv[])
...
+        database_history.open(outputPath + "/greedy_algorithm_log.txt", std::ios::app);
+
+        if (rom_online)
+        {
+            parameterPointGreedySelector->setPointResidual(residual, myid, nprocs);
+            database_history << "Residual at blast energy factor " << romOptions.blast_energyFactor << " is: " << residual << endl;
+        }
+        else
+        {
+            database_history << "ROM constucted at blast energy factor: " << romOptions.blast_energyFactor << endl;
```
chldkdtn: "constucted" is misspelled.

chldkdtn (approved):
Very nice!

kevinhkhuynh:
All tests pass except for the greedy tests which should fail.

kevinhkhuynh:
Ready for review.

rom/laghos.cpp
```diff
@@ -1334,37 +1413,40 @@ int main(int argc, char *argv[])
...
-                {
-                    MFEM_ABORT("The time step crashed!");
-                }
-                t = t_old;
-                if (!rom_online || !romOptions.hyperreduce) *S = *S_old;
-                if (rom_online) romS = romS_old;
-                if (fom_data)
+                const double dt_est = romOptions.hyperreduce ? romOper[romOptions.window]->GetTimeStepEstimateSP() : oper->GetTimeStepEstimate(*S);
+
+                //const double dt_est = oper->GetTimeStepEstimate(*S);
```
siuwuncheung: Line 1419-1421 can be removed.

rom/laghos.cpp
```diff
@@ -1727,8 +1818,81 @@ int main(int argc, char *argv[])
         }
     }
 
+    double errorIndicator = INT_MAX;
+    bool errorIndicatorComputed = false;
+    int errorIndicatorVecSize = 0;
+
     if (rom_online)
     {
+
```
siuwuncheung: Extra line

chldkdtn:
@siuwuncheung , @dylan-copeland Let's review and approve this PR soon.

rom/scripts/GREEDY.md
```diff
@@ -0,0 +1,33 @@
...
+./scripts/laghos_greedy_alg_launcher.sh laghos ... (to run locally without using srun)
+
+To use sbatch, maintaining one node throughout the algorithm. sbatch configuration
+commands can be modified within the launcher script:
+
+sbatch scripts/laghos_greedy_alg_launcher.sh srun -n 1 -p pdebug laghos ...
+
+Example command with all available options:
+
+./scripts/laghos_greedy_alg_launcher.sh laghos -m data/cube01_hex.mesh -pt 211 -tf 0.01 -build-database -writesol -romsvds -greedy-param "bef" -greedy-param-min 0.5 -greedy-param-max 2.5 -greedy-param-size 5 -greedysubsize 2 -greedyconvsize 3 -greedytol 0.1 -greedyalpha 1.05 - greedymaxclamp 2.0 -greedysamptype random -greedyerrindtype useLastLifted
```
dylan-copeland, line 20: Remove space before `greedymaxclamp`.

rom/scripts/GREEDY.md
```diff
@@ -0,0 +1,33 @@
+To use the greedy algorithm obtained from algorithm 2 of Choi et. al's
+paper "Gradient-based constrained optimization using a database
+of linear reduced-order models": https://arxiv.org/abs/1506.07849
+to build a ROM database with Laghos, use one of the following commands.
```
dylan-copeland: Since greedy algorithm could mean other algorithms, and ROM databases could be constructed in many ways, perhaps elaborate some more here that this is for building an accurate ROM database for a finite simulation parameter space.

rom/scripts/GREEDY.md
dylan-copeland: When I run this, I get
`MFEM abort: The greedy algorithm procedure has completed!`
The message is positive, but the `abort` is negative. Is this good or bad? It looks like this is at the end of the offline run, so is this the intended way of ending?

rom/laghos_rom.hpp
```diff
@@ -81,6 +120,19 @@ struct ROM_Options
...
     std::string *basename = NULL;
 
+    std::string basisIdentifier = "";
+    std::string greedyParam = "bef";
+    double greedyTol = 0.1; // error indicator tolerance for the greedy algorithm
+    double greedyAlpha = 1.05; // alpha constant for the greedy algorithm
+    double greedyMaxClamp = 2.0; // max clamp constant for the greedy algorithm
+    double greedyParamSpaceMin = 0; // min value of the greedy algorithm parameter domain
+    double greedyParamSpaceMax = 0; // max value of the greedy algorithm parameter domain
+    int greedyParamSpaceSize = 0; // size of the greedy algorithm parameter space
```
dylan-copeland: It would help to clarify what exactly is meant by `size`.

rom/laghos_rom.hpp
```diff
@@ -81,6 +120,19 @@ struct ROM_Options
 
     std::string *basename = NULL;
 
+    std::string basisIdentifier = "";
+    std::string greedyParam = "bef";
+    double greedyTol = 0.1; // error indicator tolerance for the greedy algorithm
+    double greedyAlpha = 1.05; // alpha constant for the greedy algorithm
+    double greedyMaxClamp = 2.0; // max clamp constant for the greedy algorithm
+    double greedyParamSpaceMin = 0; // min value of the greedy algorithm parameter domain
```
dylan-copeland: `parameter domain` -> `1D parameter domain`

rom/laghos_rom.hpp
```diff
@@ -81,6 +120,19 @@ struct ROM_Options
...
+    double greedyParamSpaceMax = 0; // max value of the greedy algorithm parameter domain
+    int greedyParamSpaceSize = 0; // size of the greedy algorithm parameter space
+    int greedySubsetSize = 0; // subset size of parameter points whose error indicators are checked during the greedy algorithm
+    int greedyConvergenceSubsetSize = 0; // convergence subset size for terminating the greedy algorithm
+    samplingType greedySamplingType = randomSampling; // error indicator type for the greedy algorithm
```
dylan-copeland: Wrong comment (copied and pasted).

rom/laghos_rom.cpp
```diff
@@ -3262,6 +3266,197 @@ void ROM_Operator::EvalSpaceTimeResidual_RK4(Vector const& S, Vector &f) const
...
+            readNum(romOptions.dimFv, outputPath + "/" + "rdimfv" + romOptions.basisIdentifier + "_" + to_string(errorIndicatorEnergyFraction));
+            readNum(romOptions.dimFe, outputPath + "/" + "rdimfe" + romOptions.basisIdentifier + "_" + to_string(errorIndicatorEnergyFraction));
+        }
+
+        rom_online = true;
+    }
+    else if (samplePointData != NULL)
+    {
+        CAROM::Vector* localROM = pointRequiringRelativeError.localROM.get();
+        romOptions.basisIdentifier = "_" + to_string(localROM->item(0));
```
dylan-copeland: Although it seems unlikely, could two basis identifiers be the same and cause incorrect behavior?

rom/laghos.cpp
```diff
@@ -321,14 +346,48 @@ int main(int argc, char *argv[])
...
+    NormType normtype = localmap[normtype_char];
+
+    CAROM::GreedyParameterPointSampler* parameterPointGreedySampler = NULL;
+    bool rom_calc_rel_error = false;
+
+    // If using the greedy algorithm, initialize the parameter point greedy sampler.
+    if (rom_build_database)
+    {
+        MFEM_VERIFY(!rom_offline && !rom_online && !rom_restore, "-offline, -online, -restore should be off when using -build-database");
+        parameterPointGreedySampler = BuildROMDatabase(romOptions, t_final, dt_factor, myid, outputPath, rom_offline, rom_online,  rom_calc_rel_error, greedyParam, greedyErrorIndicatorType, greedySamplingType);
```
dylan-copeland: extra space before `rom_calc_rel_error`

rom/laghos.cpp
```diff
@@ -1687,20 +1773,29 @@ int main(int argc, char *argv[])
         if(usingWindows && romOptions.parameterID == -1) outfile_twp.close();
     }
 
+    double relative_error = -1;
+
+    if (rom_build_database && rom_calc_rel_error)
+    {
+        relative_error = PrintDiffParGridFunction(normtype, myid, outputPath + "/Sol_Position" + "_" + to_string(romOptions.blast_energyFactor), x_gf);
```
dylan-copeland, line 1794: What solutions are being compared here (how were they computed and at what parameters)?

rom/laghos.cpp
dylan-copeland: I'm not sure, but it may be possible to use `ode_solver` here, if it is not used later. That is, you can change the operator and solution state for an `ODESolver`. It may be worth trying, and if it doesn't work, it's not a big deal.

rom/laghos.cpp
```diff
@@ -1334,37 +1417,40 @@ int main(int argc, char *argv[])
             const double last_dt = dt;
 
             // Adaptive time step control.
-            const double dt_est = romOptions.hyperreduce ? romOper[romOptions.window]->GetTimeStepEstimateSP() : oper->GetTimeStepEstimate(*S);
-
-            //const double dt_est = oper->GetTimeStepEstimate(*S);
-            //cout << myid << ": dt_est " << dt_est << endl;
-            if (dt_est < dt)
+            if (!rom_build_database || last_step)
```
dylan-copeland, line 1436: What happens if `rom_build_database == true`, `last_step == true`, and the timestep is reset here? Are there any issues with it recomputing the final timestep with a smaller timestep size? Since it would require more than 1 step, it would no longer be the final step. This is a very rare situation, so maybe it has not been tested. If it is a complication that should be resolved in a later PR (if it becomes an issue), then there could be just an error thrown if the timestep is reduced in the `rom_build_database` case.

rom/laghos.cpp
```diff
@@ -1727,8 +1822,81 @@ int main(int argc, char *argv[])
...
 
+    double errorIndicator = INT_MAX;
+    bool errorIndicatorComputed = false;
+    int errorIndicatorVecSize = 0;
+
     if (rom_online)
     {
+
+        // If using the greedy algorithm, calculate the error indicator using the FOM lifted during
+        // the second to last step compared against the FOM lifted at the last step.
```
dylan-copeland: Is this comment meant for the `useLastLiftedSolution` case?

rom/laghos.cpp
```diff
@@ -1727,8 +1822,81 @@ int main(int argc, char *argv[])
...
+            if (romOptions.greedyErrorIndicatorType == useLastLiftedSolution)
+            {
+                if (converged)
+                {
+                    Vector errorIndicatorVec = Vector(lastLiftedSolution.Size());
+                    subtract(lastLiftedSolution, *S, errorIndicatorVec);
+
+                    errorIndicator = errorIndicatorVec.Norml2();
+                    errorIndicatorVecSize = errorIndicatorVec.Size();
+                }
```
dylan-copeland, line 1861: What happens in the `!converged` case, when `errorIndicator == INT_MAX`?

rom/laghos.cpp
```diff
@@ -1289,34 +1349,57 @@ int main(int argc, char *argv[])
                 if (myid == 0)
                     cout << "ROM online at t " << t << ", dt " << dt << ", romS norm " << romS.Norml2() << endl;
 
+                if (rom_build_database && !std::isfinite(romS.Norml2()))
```
dylan-copeland, line 1367: What would cause `romS` to not be finite?

rom/scripts/GREEDY.md
kevinhkhuynh: Yes, it's a bit tough because without being able to run Laghos simulations multiple times within one command, we have to run it multiple times in a script. The only way the script would know when to stop running new Laghos simulations would be either to check for some file/message, which is a potential option, or through the exit code (which is what the MFEM abort is for).

rom/laghos_rom.cpp
kevinhkhuynh: Yes. Two identical basis identifiers would be a bug and should not happen. The greedy algorithm has checks to prevent the same point being sampled twice.

rom/laghos.cpp
kevinhkhuynh: The greedy algorithm told us to make a new ROM at point B because the residual was high, and it is predicting that the relative error at point B will be greater than 0.1 (or some other threshold). So we create a new ROM at point B. We now want to check whether the greedy algorithm made a correct prediction.

We use the ROM made at point A to run an online_sim at point B. We compare the lifted ROM solution at point B to the FOM solution at point B (which we have already saved from a previous run) to get the relative_error.

If the relative error was lower than 0.1, then we actually didn't need to make a ROM at point B. The greedy algorithm needs to make some adjustments.

rom/laghos.cpp
kevinhkhuynh: Ah, I didn't think about this. it looks like we would just go to another timestep and the second to last timestep solution would be thrown away and recalculated, and there shouldn't be a problem, but this hasn't been tested yet.

rom/laghos.cpp
kevinhkhuynh: Yes, I'll clarify.

rom/laghos.cpp
kevinhkhuynh: The greedy algorithm takes this information and sets the error indicator to INT_MAX, meaning we definitely need to build a ROM near that point. This was intended.

rom/laghos.cpp
kevinhkhuynh: If MFEM does not converge for some reason, the run will take forever. This can happen if the ROM is too far away from the point we are now trying to find a residual at. This kills the run at the step after it does not converge, since romS will be full of Nan's.

rom/scripts/GREEDY.md
dylan-copeland: I see, this is a tricky case for which the original Laghos driver was not designed. Using `MFEM_Abort` does seem convenient, but it can be confusing if one usually expects it to mean a failure. Also, the memory is not freed at the end, which I suppose doesn't really matter but is not good style. Maybe a cleaner way to end would be for `SaveROMDatabase` to return the exit code, so that the memory is still freed at the end of `main` and then the code is returned by `main` on a proper exit. Checking a file for a message is another possibility, but the return code seems simplest.

rom/laghos.cpp
dylan-copeland: This makes sense. A comment briefly explaining that would be helpful.

rom/laghos_rom.cpp
dylan-copeland: My question is actually for multidimensional parameters, which are not implemented (yet) in Laghos. The issue is that it only uses `item(0)` for the identifier. For 1D parameters like `bef` that is fine, but in general I suppose it should depend on all entries, to avoid different `localROM` values having the same `basisIdentifier`.

rom/scripts/GREEDY.md
```diff
@@ -0,0 +1,33 @@
...
+
+Hyperreduction is not supported!
```
dylan-copeland: Why not? Would the online simulations be faster with hyperreduction, or does it not matter since FOM simulations are needed anyway?

rom/laghos_rom.hpp
```diff
@@ -81,6 +120,19 @@ struct ROM_Options
 
     std::string *basename = NULL;
 
+    std::string basisIdentifier = "";
+    std::string greedyParam = "bef";
+    double greedyTol = 0.1; // error indicator tolerance for the greedy algorithm
```
dylan-copeland: In GREEDY.md this is described as a relative tolerance, and my understanding is that errors in Laghos used by the libROM greedy algorithm are always relative errors. It may be clearer to add to the comment here that this is for relative error.

rom/laghos_rom.hpp
```diff
@@ -38,6 +39,19 @@ enum offsetStyle
...
+    randomSampling,
+    latinHypercubeSampling
+};
+
+enum errorIndicatorType
+{
+    useLastLiftedSolution,
+    varyTimeStep,
+    varyBasisSize
+};
```
dylan-copeland, line 52: Are there only tests for `useLastLiftedSolution`? The `vary*` options are not documented in `GREEDY.md`, and I'm not sure if they are really supported. For example, I don't see where `dt_factor` is ever changed from 1.

dylan-copeland:
The result of running `sbatch tests/runRegressionTests.sh -t -i "sedov-blast"` has
sedov-blast-greedy-use_database_online: PASS
sedov-blast-greedy-use_database_online-parallel: FAIL
with the error
unable to open file: name = 'run/sedov-blast-greedy-parallel/greedy_algorithm_data_157531.000002'
Should this pass with the `-t` option?

rom/scripts/GREEDY.md
kevinhkhuynh: This is coming in the next PR.

rom/laghos_rom.hpp
kevinhkhuynh: I will add tests for the other two. Currently, the user should input their -dt_factor option they want to test the greedy algorithm with (should be different than 1), then the greedy algorithm will set the dt_factor to 1 for one of the runs and use the user's dt_factor for the other in the BuildROMDatabase function in laghos_rom.cpp.

kevinhkhuynh: If this seems a bit backwards to how the dtFactor option should be used, I can instead add another greedyDt option

kevinhkhuynh: Ah never mind. Just noticed something wrong about it. Modifying to have a greedyDt option.

kevinhkhuynh:
> 
> 
> The result of running `sbatch tests/runRegressionTests.sh -t -i "sedov-blast"` has
> sedov-blast-greedy-use_database_online: PASS
> sedov-blast-greedy-use_database_online-parallel: FAIL
> with the error
> unable to open file: name = 'run/sedov-blast-greedy-parallel/greedy_algorithm_data_157531.000002'
> Should this pass with the `-t` option?

The tests seem to be passing for me with the '-t' option. Would you mind trying again with an updated libROM and Laghos?

rom/laghos.cpp
kevinhkhuynh: Ended up just removing the varyTimeStep, since it was didn't perform well as an error indicator anyways. Added a test for varyBasisSize.

dylan-copeland: My question is for any `errorIndicatorType`, including `useLastLiftedSolution`. In general, my understanding is that the final timestep needs to be run, in order to evaluate the residual at the final timestep. However, the new run (using the new ROM) could possibly reduce the timestep, which means the final timestep would not be the same. It looks like the answer is that is not possible, since `!rom_build_database == false` will prevent the timestep reduction here. Is `dt` input when building a database?

dylan-copeland:
@kevinhkhuynh With the latest versions of Laghos and libROM, the regression tests are passing. The run time for the regression tests is significantly longer due to this PR, but it is important to have thorough testing.

rom/laghos.cpp
kevinhkhuynh: Oops. I wrote the previous comment in the wrong thread.

dylan-copeland: Actually, I think the final `dt` cannot be prescribed or ensured to be a certain value, but the final residual can still be used.

dylan-copeland (approved):

siuwuncheung (approved):

PR #138, opened 2021-03-14
SNS implementation

siuwuncheung:
This branch is built for a new implementation of the SNS nonlinear model reduction. As the position is governed by the velocity in a linear sense, only nonlinear model reduction for velocity and energy is done. 

A new flag `-romsns` is used to enable/disable SNS. Note that this flag is independent of an existing flag `-romsrhs`, which is always set to be true in the regression tests and numerical results in the LaghosROM paper, irrespective of using snapshot SVD or SNS or nonlinear term basis. Since `-romsrhs` no longer controls the formation of nonlinear term basis and is always used, we should safely remove it in the future.

In the offline phase, `-romsns` controls whether or not to sample nonlinear term snapshots. This is originally controlled by `-romsrhs`. 

In the online phase, with the flag `-romsns`, the nonlinear term bases `basisFv` and `basisFe` are obtained from multiplying `Mv` to `basisV` and `Me` to `basisE` respectively, instead of doing SVD separately.  

Some changes in the procedure of computing the reduced matrices are made. Following the LaghosROM paper, the ROM time marching is composed of three parts. 

1. Solving for the ROM nonlinear terms coordinates
2. Galerkin projection onto ROM solution bases. 
3. Inverting the ROM mass matrices. 

Note that step 2 is not necessary if SNS type I, i.e. not extending the nonlinear term bases, is used, while step 3 is not necessary if induced Gram Schmidt or SNS type I is used. If induced Gram Schmidt is used, the velocity and energy bases are changed, and the Galerkin projection in Step 2 has to be after the induced Gram Schmidt is done. The overall hyper-reduction procedure in the current implementation is listed as follows: 

1. If SNS, multiply the mass matrices to the solution bases to obtain the nonlinear term bases. Otherwise, read the bases. 
2. Perform DEIM on the nonlinear term bases to get the sampling matrices (i.e. for step 1 above)
3. In time marching, given a new time window, 
- perform induced Gram Schmidt (omit if not Gram Schmidt)
- act Galerkin projection on the sampling matrices (i.e. for step 2 above, omit if SNS type I)
- in each step, inverting the ROM mass matrices (i.e. for step 3 above, omit if Gram-Schmidt is done or SNS type I)

After this PR is merged, the following should be considered: 

- Remove `useReducedM` and `RHSbasis` (and the flag `-romsrhs` in the regression tests) if they are no longer used.
- Change the default values of some frequently used tags, e.g. `-romos`, `-romsns`, `-twp`.

### DIFF
--- a/rom/laghos_rom.cpp
+++ b/rom/laghos_rom.cpp
@@ -281,9 +281,10 @@ CAROM::Matrix* ReadBasisROM(const int rank, const std::string filename, const in
     return basisCopy;
 }
 
-CAROM::Matrix* MultM(hydrodynamics::LagrangianHydroOperator *lhoper, const int var, const int N, const CAROM::Matrix* A)
+CAROM::Matrix* MultBasisROM(const int rank, const std::string filename, const int vectorSize, const int rowOS, int& dim, hydrodynamics::LagrangianHydroOperator *lhoper, const int var)
 {
-    CAROM::Matrix* S = new CAROM::Matrix(A->numRows(), std::min(N, A->numColumns()), A->distributed());
+    CAROM::Matrix* A = ReadBasisROM(rank, filename, vectorSize, rowOS, dim);
+    CAROM::Matrix* S = new CAROM::Matrix(A->numRows(), A->numColumns(), A->distributed());
     Vector Bej(A->numRows());
     Vector MBej(A->numRows());
 
@@ -1404,8 +1405,8 @@ void ROM_Basis::ReadSolutionBases(const int window)
     }
     else
     {
-        basisFv = MultM(lhoper, 1, rdimfv, basisV);
-        basisFe = MultM(lhoper, 2, rdimfe, basisE);
+        basisFv = MultBasisROM(rank, basename + "/" + ROMBasisName::V + std::to_string(window), tH1size, 0, rdimfv, lhoper, 1);
+        basisFe = MultBasisROM(rank, basename + "/" + ROMBasisName::E + std::to_string(window), tL2size, 0, rdimfe, lhoper, 2);
     }
 }
 
--- a/rom/laghos_rom.cpp
+++ b/rom/laghos_rom.cpp
@@ -16,7 +16,7 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const& 
     const bool sampleX = generator_X->isNextSample(t);
 
     Vector dSdt;
-    if (sampleF)
+    if (!sns)
     {
         dSdt.SetSize(S.Size());
         lhoper->Mult(S, dSdt);
@@ -91,7 +91,7 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const& 
             }
         }
 
-        if (sampleF)
+        if (!sns)
         {
             MFEM_VERIFY(gfH1.Size() == H1size, "");
             for (int i=0; i<H1size; ++i)
@@ -135,7 +135,7 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const& 
             generator_E->computeNextSampleTime(E.GetData(), dEdt.GetData(), t);
         }
 
-        if (sampleF)
+        if (!sns)
         {
             MFEM_VERIFY(gfL2.Size() == L2size, "");
             for (int i=0; i<L2size; ++i)
@@ -174,7 +174,7 @@ void ROM_Sampler::Finalize(const double t, const double dt, Vector const& S, Arr
         if (!useXV) generator_X->writeSnapshot();
         if (!useVX) generator_V->writeSnapshot();
         generator_E->writeSnapshot();
-        if (sampleF)
+        if (!sns)
         {
             generator_Fv->writeSnapshot();
             generator_Fe->writeSnapshot();
@@ -185,7 +185,7 @@ void ROM_Sampler::Finalize(const double t, const double dt, Vector const& S, Arr
         if (!useXV) generator_X->endSamples();
         if (!useVX) generator_V->endSamples();
         generator_E->endSamples();
-        if (sampleF)
+        if (!sns)
         {
             generator_Fv->endSamples();
             generator_Fe->endSamples();
@@ -212,7 +212,7 @@ void ROM_Sampler::Finalize(const double t, const double dt, Vector const& S, Arr
         BasisGeneratorFinalSummary(generator_E, energyFraction, cutoff[2]);
         PrintSingularValues(rank, basename, "E", generator_E);
 
-        if (sampleF)
+        if (!sns)
         {
             cout << "Fv basis summary output: ";
             BasisGeneratorFinalSummary(generator_Fv, energyFraction, cutoff[3]);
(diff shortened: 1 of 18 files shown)


siuwuncheung:
@chldkdtn @dylan-copeland @kevinhkhuynh This PR is ready for review. The regression tests I run yesterday all passed. To play safe we shall do LGTM. I am trying to reproduce the results in paper (which has not serial ROM merged) with the most recent commit in this PR, but Quartz is not responding. After your approval and the validation in the results, we can proceed to merge to the rom branch.

rom/laghos_rom.hpp
```diff
@@ -683,8 +686,9 @@ class ROM_Operator : public TimeDependentOperator
 
     mutable double dt_est_SP = 0.0;
 
-    bool useReducedMv = false;  // TODO: remove this?
-    bool useReducedMe = false;  // TODO: remove this?
+    bool sns1 = false;
```
dylan-copeland: The meaning of SNS Type 1 should be documented somewhere.

rom/laghos_rom.hpp
```diff
@@ -65,6 +65,7 @@ struct ROM_Options
     bool staticSVD = false; // true: use StaticSVD
     bool useOffset = false; // if true, sample variables minus initial state as an offset
     bool RHSbasis = false; // if true, use bases for nonlinear RHS terms without mass matrix inverses applied
+    bool SNS = false; // if true, use sns
```
dylan-copeland: We know what SNS is, but maybe for users it should be explained briefly, maybe with a reference to the SNS paper.

rom/laghos_rom.cpp
```diff
@@ -2272,28 +2260,21 @@ void ROM_Operator::UndoInducedGramSchmidt(const int var, Vector &S, bool keep_da
...
-    InducedGramSchmidt(2, S); // energy
-    basis->ComputeReducedMatrices();
-    if (useReducedMv)
-    {
-        ComputeReducedMv();
-        ComputeReducedMe();
-    }
-
-    if (hyperreduce)
+    if (useGramSchmidt && !sns1)
```
dylan-copeland, line 2264: Now this function is called even if `useGramSchmidt` is false. Would it be more accurate to rename the function as something like `SetupReducedMatrices`?

rom/laghos_rom.cpp
```diff
@@ -281,11 +281,37 @@ CAROM::Matrix* ReadBasisROM(const int rank, const std::string filename, const in
     return basisCopy;
 }
 
+CAROM::Matrix* MultBasisROM(const int rank, const std::string filename, const int vectorSize, const int rowOS, int& dim, hydrodynamics::LagrangianHydroOperator *lhoper, const int var)
```
dylan-copeland: Long line --> 2 lines

kevinhkhuynh:
Is it okay if I push a tiny one-line change to the 'make setup' script here? hypre-2.11.2 no longer exists, it seems LLNL has deleted it. So we have to switch to cloning the github hypre repo for LGTM to be able to run.

rom/laghos_rom.cpp
```diff
@@ -1379,6 +1360,17 @@ void ROM_Basis::ReadSolutionBases(const int window)
         MFEM_VERIFY(basisX->numRows() == tH1size, "");
         basisV = basisX;
     }
+
+    if (!use_sns)
```
dylan-copeland: It would be easier to read if the cases were reversed, with `if (use_sns)`.

rom/merge.cpp
```diff
@@ -438,9 +442,9 @@ int main(int argc, char *argv[])
             }
         }
 
-        MFEM_VERIFY(dimX == dimV, "Different sizes for X and V");
-        MFEM_VERIFY(dimFv == dimV, "Different sizes for V and Fv");
-        MFEM_VERIFY(dimFe == dimE, "Different sizes for E and Fe");
+        //MFEM_VERIFY(dimX == dimV, "Different sizes for X and V");
```
dylan-copeland: Shouldn't this be kept? It checks whether the X and V are in the same space.

rom/laghos_rom.hpp
siuwuncheung: @dylan-copeland A comment is added and the corresponding equation in the SNS paper is mentioned.

rom/laghos_rom.hpp
siuwuncheung: The arXiv ID is of the SNS paper is added. Once our LaghosROM paper is available online, we can use that as reference.

rom/laghos_rom.cpp
siuwuncheung: @dylan-copeland Thanks for the suggestion. The cases are reversed.

dylan-copeland:
There are some compiler warnings that look easy to fix.

rom/merge.cpp
siuwuncheung: I don't remember changing this part in this PR. I think this is commented long ago, and all these three verifies are not necessarily true. If we use offset, then the true case should be `dimFv = dimV + 1` and `dimFe = dimE + (V0 ~= 0)`. On the other hand, if we don't use offset and the initial velocity (in this window) is zero, then the true case should be `dimX = dimV + 1` and `dimFe = dimE + 1`.

rom/laghos_rom.cpp
siuwuncheung: The routine is now renamed as `ApplyHyperreduction`.

rom/laghos_rom.cpp
siuwuncheung: It is now broken into 2 lines.

rom/merge.cpp
dylan-copeland: I think `dim` is the number of DOFs, and `snapshotSize` is the dimension you are referring to which can differ, so these verifications should be valid. Could you try running the regression tests with these uncommented to see if there is a case where they fail? The PR shows this was changed in this branch.

siuwuncheung:
> Is it okay if I push a tiny one-line change to the 'make setup' script here? hypre-2.11.2 no longer exists, it seems LLNL has deleted it. So we have to switch to cloning the github hypre repo for LGTM to be able to run.

@kevinhkhuynh Please feel free to do the change here. As far as this will be the next branch to be merged, i suppose it should be fine.

siuwuncheung:
> There are some compiler warnings that look easy to fix.

The warning has just been fixed.

kevinhkhuynh:
LGTM

kevinhkhuynh:
Not sure if everyone saw it, but pipeline passed.

rom/merge.cpp
siuwuncheung: @dylan-copeland Thanks. You're right about those lines and I have uncommented them.

dylan-copeland (approved):
Thanks @siuwuncheung!

chldkdtn (approved):

kevinhkhuynh (approved):

chldkdtn:
Looks like it is ready to merge, right?

PR #142, opened 2021-03-30
A new ROM implementation of Rayleigh-Taylor instability problem

siuwuncheung:

The following results are generated using the main directory which mirrors the most recent `master`.

Density at `t = 4.0`

master_density

Velocity at `t = 4.0`

master_velocity

Energy at `t = 4.0`

master_energy


siuwuncheung:
All regression tests passed up to this point. 

Density at `t = 4.0`
density

Velocity at `t = 4.0`
velocity

Energy at `t = 4.0`
energy

siuwuncheung:
After a change in boundary condition, regression tests are broken as expected, but with negligible error.

Density at `t = 4.0`
density

Velocity at `t = 4.0`
velocity

Energy at `t = 4.0`
energy

laghos.cpp
```diff
@@ -605,7 +613,27 @@ int main(int argc, char *argv[])
    BlockVector S_old(S);
    long mem=0, mmax=0, msum=0;
    int checks = 0;
-
+   //   const double internal_energy = hydro.InternalEnergy(e_gf);
```
dylan-copeland, line 616: If this commented block of code should be kept for convenience, it would be better to use a `bool` to decide whether to do the output rather than commenting out.

laghos.cpp
```diff
@@ -663,15 +691,24 @@ int main(int argc, char *argv[])
...
          {
             const double sqrt_norm = sqrt(norm);
+
             cout << std::fixed;
             cout << "step " << std::setw(5) << ti
                  << ",\tt = " << std::setw(5) << std::setprecision(4) << t
                  << ",\tdt = " << std::setw(5) << std::setprecision(6) << dt
                  << ",\t|e| = " << std::setprecision(10) << std::scientific
                  << sqrt_norm;
+            //  << ",\t|IE| = " << std::setprecision(10) << std::scientific
```
dylan-copeland, line 706: Again, use a condition rather than commenting out.

laghos.cpp
siuwuncheung: These changes made to the main directory are actually merged from the `master` branch. I guess we should better not make any changes so that when we merge from the `master` in the future, it would not cause any conflicts.

laghos.cpp
siuwuncheung: These changes made to the main directory are actually merged from the `master` branch. I guess we should better not make any changes so that when we merge from the `master` in the future, it would not cause any conflicts.

laghos.cpp
dylan-copeland: Yes, I see now that it is in master, which is surprising. I agree to keep it, to avoid conflicts.

rom/laghos.cpp
```diff
@@ -611,20 +614,23 @@ int main(int argc, char *argv[])
...
     {
         {
-            Array<int> ess_bdr(pmesh->bdr_attributes.Max()), tdofs1d;
+            Array<int> ess_bdr(pmesh->bdr_attributes.Max()), dofs_marker, dofs_list;
             for (int d = 0; d < pmesh->Dimension(); d++)
             {
                 // Attributes 1/2/3 correspond to fixed-x/y/z boundaries, i.e., we must
                 // enforce v_x/y/z = 0 for the velocity components.
-                ess_bdr = 0;
+                ess_bdr = 0;
```
dylan-copeland: Extra space. Apparently `make style` does not eliminate this.

dylan-copeland:
The test `./laghos -p 7 -m data/rt2D.mesh -tf 4 -rs 1 -ok 4 -ot 3 -vis -pa` works in parallel (e.g. `mpirun -n 4`) for the main version, but for the `rom/` version the timestep size goes to zero. @siuwuncheung have you looked into the parallel case?

siuwuncheung:
> The test `./laghos -p 7 -m data/rt2D.mesh -tf 4 -rs 1 -ok 4 -ot 3 -vis -pa` works in parallel (e.g. `mpirun -n 4`) for the main version, but for the `rom/` version the timestep size goes to zero. @siuwuncheung have you looked into the parallel case?

I have been testing the serial runs only. I will look into the parallel runs.

siuwuncheung:
> > The test `./laghos -p 7 -m data/rt2D.mesh -tf 4 -rs 1 -ok 4 -ot 3 -vis -pa` works in parallel (e.g. `mpirun -n 4`) for the main version, but for the `rom/` version the timestep size goes to zero. @siuwuncheung have you looked into the parallel case?
> 
> I have been testing the serial runs only. I will look into the parallel runs.

The parallel runs have been fixed.

dylan-copeland (approved):
Thanks @siuwuncheung I verified that the parallel run works. The failing Taylor-Green regression tests are acceptable to me, although the cause is unclear, as I verified that the errors and speed-up for the Taylor-Green test in our paper have not changed significantly.

chldkdtn (approved):

kevinhkhuynh (approved):

PR #148, opened 2021-04-29
Cleanup the ROM branch

siuwuncheung:
Cleaning up the offline parameter file I/O after introducing the parametric Rayleigh-Taylor problem.


dylan-copeland (approved):
Thanks @siuwuncheung! I confirmed that the triple-point problem works as before now. The triple-point regression tests all fail, but all other regression tests pass, which is clearly due to the change in that test.

chldkdtn (approved):

PR #149, opened 2021-05-10
New local ROM approaches for Rayleigh-Taylor instability

siuwuncheung:
Add an implementation of using penetration distance and using the parameter-time rather than physical time as the indicator the trust-region of local ROM for 2D Rayleigh-Taylor instability.


siuwuncheung:
@chldkdtn @dylan-copeland @kevinhkhuynh Ready for review. Undergoing regression tests.

rom/laghos.cpp
```diff
@@ -1483,7 +1513,21 @@ int main(int argc, char *argv[])
...
                 timeLoopTimer.Stop();
                 samplerTimer.Start();
-                sampler->SampleSolution(t, last_dt, *S);
+                // 2D Rayleigh-Taylor penetration distance
+                double real_pd = -1.0;
+                if (problem == 7)
+                {
+                    if (romOptions.indicatorType == penetrationDistance)
+                    {
+                        double proc_pd = (pd2_vdof > 0) ? -(*S)(pd2_vdof) : 0.0;
```
dylan-copeland: Should this be `pd2_vdof >= 0`?

rom/laghos.cpp
```diff
@@ -1969,12 +2046,13 @@ int main(int argc, char *argv[])
         // 2D Rayleigh-Taylor penetration distance
         if (problem == 7 && fom_data)
         {
-            double my_pd[2], pd_max[2];
-            my_pd[0] = (pd1_vdof > 0) ?  (*S)(pd1_vdof) : 0.0;
-            my_pd[1] = (pd2_vdof > 0) ? -(*S)(pd2_vdof) : 0.0;
-            MPI_Reduce(my_pd, pd_max, 2, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
+            double proc_pd[2], real_pd[2];
+            proc_pd[0] = (pd1_vdof > 0) ?  (*S)(pd1_vdof) : 0.0;
```
dylan-copeland: Should this be `pd1_vdof >= 0`?

rom/laghos_rom.cpp
```diff
@@ -2277,6 +2288,28 @@ void ROM_Basis::readSP(ROM_Options const& input, const int window)
     }
 }
 
+void ROM_Basis::writePD(const int id, const int window) const
```
dylan-copeland: Would it be clearer to call this function `writePDweights`?

dylan-copeland:
@siuwuncheung Are there any new tests? I don't see any changes to the regression tests. Should there be a short regression test for maintaining the new capability? Also, is there a longer, more interesting test we can try? Does it work in parallel?

rom/laghos_utils.cpp
```diff
@@ -40,15 +40,16 @@ int WriteOfflineParam(int dim, double dt, ROM_Options& romOptions,
...
     if (printStatus)
     {
+        std::ofstream opout(paramfile, std::fstream::app);
         opout << romOptions.parameterID << " ";
         opout << romOptions.rhoFactor << " ";
         opout << romOptions.blast_energyFactor << " ";
         opout << romOptions.atwoodFactor << endl;
         opout.close();
     }
+    return 0;
```
dylan-copeland: What is the purpose of the return value? It seems this function, as well as `VerifyOfflineParam`, could be `void`.

siuwuncheung: Yes it is meaningless and should be removed.

rom/laghos_rom.cpp
siuwuncheung: I agree. I have changed the name of the function.

rom/laghos.cpp
siuwuncheung: Absolutely.

siuwuncheung:
> @siuwuncheung Are there any new tests? I don't see any changes to the regression tests. Should there be a short regression test for maintaining the new capability? Also, is there a longer, more interesting test we can try? Does it work in parallel?

- I will add new short regression tests for this capability. 
- The command line options and the results for long time simulation with final time is 4.5 is tabulated on Confluence. 
- The penetration distance is in the context of Rayleigh-Taylor instability only, but in general problems there can be a small set of easy-to-evaluate functional which marks or is highly correlated to the position of the shock, the same idea should apply (probably the Sedov Blast explosion or the Riemann problem). 
- I think the merge phase does not work in parallel and therefore the online phase either, but I may be wrong. In that case we need to fix that, especially if we want to use higher order discretizations on fine meshes to show a good speed up.

chldkdtn (approved):

chldkdtn:
@siuwuncheung what is the status of this PR?

siuwuncheung:
> @siuwuncheung what is the status of this PR?

There are some new capabilities that I want to add to enable parametric distance windowing for long-time simulation. Right now the offset problem is resolved and some file I/O settings are being added. After that I think this PR should be reviewed or merge, or otherwise it will grow too large and complicated.

kevinhkhuynh (approved):

rom/laghos.cpp
```diff
@@ -1727,14 +1817,30 @@ int main(int argc, char *argv[])
                     {
                         romOper[romOptions.window]->ApplyHyperreduction(romS);
                     }
+
+                    if (problem == 7 && romOptions.indicatorType == penetrationDistance)
+                    {
+                        std::string pd_weight_outPath = testing_parameter_outputPath + "/pd_weight" + to_string(romOptions.window);
```
dylan-copeland: Lines 1823-1833 are the same as lines 1271-1281 except for the window number. This could be made into a function, to avoid code duplication.

rom/laghos.cpp
```diff
@@ -931,14 +941,20 @@ int main(int argc, char *argv[])
 
     // 2D Rayleigh-Taylor penetration distance
     int pd1_vdof = -1, pd2_vdof = -1;
```
dylan-copeland, line 944: A brief comment here would be helpful, explaining that these are position/velocity DOFs of a particular point hard-coded for problem 7.

rom/laghos.cpp
```diff
@@ -1576,19 +1617,51 @@ int main(int argc, char *argv[])
...
-                        sampler->SampleSolution(RKTime[RKidx], last_dt, RKStages[RKidx]);
-                        if (samplerLast) samplerLast->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
+                        if (problem == 7)
+                        {
+                            // 2D Rayleigh-Taylor penetration distance
+                            if (romOptions.indicatorType == penetrationDistance)
+                            {
+                                real_pd = -1.0;
+                                double proc_pd = (pd2_vdof >= 0) ? -RKStages[RKidx](pd2_vdof) : 0.0;
+                                MPI_Reduce(&proc_pd, &real_pd, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
```
dylan-copeland, line 1629: Should this be `MPI_Allreduce`? Using `MPI_Reduce` just computes the result `real_pd` on the root process, but isn't it needed on all processes?

rom/laghos.cpp
```diff
@@ -658,9 +667,10 @@ int main(int argc, char *argv[])
 
     std::string offlineParam_outputPath = outputPath + "/offline_param" + romOptions.basisIdentifier + ".csv";
     romOptions.offsetType = getOffsetStyle(offsetType);
+    romOptions.indicatorType = getlocalROMIndicator(indicatorType);
     if (rom_online)
     {
-        std::string filename = solution_outputPath + "/ROMsol/romS_1";
+        std::string filename = testing_parameter_outputPath + "/ROMsol/romS_1";
```
dylan-copeland, line 673: I realize `romS_1` was used before this PR, but I'm not sure why we don't check for `romS_0` in case there is only one window.

rom/laghos_rom.hpp
```diff
@@ -602,6 +623,13 @@ class ROM_Basis
     void writeSP(ROM_Options const& input, const int window = 0) const;
     void readSP(ROM_Options const& input, const int window = 0);
 
+    void writePDweights(const int id, const int window = 0) const;
+
+    double GetOffsetX(const int idx)
```
dylan-copeland: This could be a `const` function like the other `Get` functions.

rom/laghos_rom.cpp
```diff
@@ -471,7 +479,7 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, MPI_Comm comm_, MPI_Comm rom_com_
 
     if (offsetInit)
     {
-        std::string path_init = basename + "/ROMoffset" + input.basisIdentifier + "/init";
+        std::string path_init = basename + "/ROMoffset" + input.basisIdentifier + "/init"; // TODO: think about parametric case
```
dylan-copeland: Is this still an issue? If so, what remains to be resolved?

rom/laghos_rom.cpp
```diff
@@ -2149,42 +2157,47 @@ void ROM_Basis::computeWindowProjection(const ROM_Basis& basisPrev, ROM_Options
...
-        initXsp->write(basename + "/" + "initXsp" + "_" + to_string(window));
-        initVsp->write(basename + "/" + "initVsp" + "_" + to_string(window));
-        initEsp->write(basename + "/" + "initEsp" + "_" + to_string(window));
+        initXsp->write(testing_parameter_basename + "/" + "initXsp" + "_" + to_string(window));
+        initVsp->write(testing_parameter_basename + "/" + "initVsp" + "_" + to_string(window));
+        initEsp->write(testing_parameter_basename + "/" + "initEsp" + "_" + to_string(window));
     }
 
-    if (window > 0)
+    if (window > 0) // TODO: do not output multiple times
```
dylan-copeland, line 2204: Does this mean these matrices are written to file for each testing parameter, although they are always the same?

rom/laghos_rom.cpp
```diff
@@ -54,6 +54,7 @@ void ROM_Sampler::SampleSolution(const double t, const double dt, Vector const&
         if (writeSnapshots && addSample)
         {
             tSnapX.push_back(t);
+            if (pd >= 0.0) pdSnap.push_back(pd);
```
dylan-copeland, line 57: In multiple places, there is logic like this based on whether `pd >= 0`, with negative `pd` indicating that penetration distance is not used. I suppose this is designed just for the RT problem 7, in which `pd` is always non-negative. In general, could this be negative, or is that unrealistic? Should the logic be generalized to use some flag instead of checking `pd >= 0`, or should this work for the problems we are interested in?

siuwuncheung: I believe that's unrealistic (in physical meanings) since the initial velocity drives the flow in the anti-clockwise and the gravitation drives it downward. We have also seen in the experiments that the penetration distance at time `t` with Atwood number `A` closely fits with the theoretical finding `C*A*t^2` for some constant C. Right now the only application of an extra indicator in Laghos is the penetration distance (or `A*t^2` of the Rayleigh Taylor instability, but if we want to generalize to some other indicators in other experiments, we can do the corresponding changes easily. I lean to making these changes whenever necessary, but I can do it in this PR, or I can add comments so that our team will understand how to make the changes later.

rom/laghos_rom.cpp
siuwuncheung: Yes, that is right. I have not changed this yet. It is not an urgent one. I can do it in this PR or the next.

rom/laghos_rom.cpp
siuwuncheung: This is now fixed.

rom/laghos_rom.hpp
siuwuncheung: Made to `const`

rom/laghos.cpp
siuwuncheung: I think the existence of `romS_0` is legit in the hyperreduce time marching phase since it is produced in the preparation phase. Anything beyond 0 is prohibited and 1 is used.

rom/laghos.cpp
siuwuncheung: I believe  `MPI_Reduce` suffices since the use of `real_pd` here is to record the FOM penetration distance in the offline phase, which will be written into a file for merge phase (if it is parametric problem) and online phase (for the determination of window change) and only the master node needs a record.

rom/laghos.cpp
siuwuncheung: Comment is added.

rom/laghos.cpp
siuwuncheung: Changes have been made accordingly.

rom/laghos.cpp
dylan-copeland: Thanks for clarifying. I see now that it is used only by the root node. It is still not clear to me how the `MFEM_VERIFY` does not fail for processes other than root, which should have `real_pd == -1.0`. Have you tested this in parallel? Also, I think it would be a little better to set `double real_pd = -1.0;` in the declaration of this variable, so it is not uninitialized in some case.

rom/laghos_rom.cpp
dylan-copeland: I agree it is fine for now.

rom/tests/runRegressionTests.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 #SBATCH -N 1
-#SBATCH -t 1:00:00
-#SBATCH -p pdebug
+#SBATCH --exclusive
+#SBATCH -t 2:00:00
```
dylan-copeland: There was a very slow test that took about 1 hour, but @kevinhkhuynh removed it, and now the tests all run in under 1 hour. The time limit `-t 1:00:00` could be restored here.

siuwuncheung: Restored.

rom/laghos.cpp
dylan-copeland: I noticed that you changed some `MPI_Reduce` calls to `MPI_Allreduce`, but there are still some left (related to `real_pd`).

rom/laghos_rom.cpp
```diff
@@ -479,13 +479,14 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, MPI_Comm comm_, MPI_Comm rom_com_
...
-        std::string path_init = testing_parameter_basename + "/ROMoffset" + input.basisIdentifier + "/init"; 
+        std::string path_init = testing_parameter_basename + "/ROMoffset" + input.basisIdentifier + "/init";
 
-        if (input.offsetType == useInitialState)
+        if (input.offsetType == useInitialState && (input.restore || input.window > 0))
         {
             cout << "Reading: " << path_init << endl;
 
-            // Read offset in the online phase of initial mode
+            // Read initial offset in the resotre phase or online phase
```
dylan-copeland: typo: restore

rom/laghos_rom.cpp
```diff
@@ -496,15 +497,15 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, MPI_Comm comm_, MPI_Comm rom_com_
...
+            // Read window dependent offsets in the restore phase or in the online phase
             initX->read(path_init + "X" + std::to_string(input.window));
             initV->read(path_init + "V" + std::to_string(input.window));
             initE->read(path_init + "E" + std::to_string(input.window));
-
             cout << "Read init vectors X, V, E with norms " << initX->norm() << ", " << initV->norm() << ", " << initE->norm() << endl;
         }
         else if (input.offsetType == interpolateOffset)
         {
+            // Interpoate and save window dependent offsets in the online phase
```
dylan-copeland: Interpolate

rom/laghos_rom.cpp
```diff
@@ -546,7 +549,7 @@ ROM_Basis::ROM_Basis(ROM_Options const& input, MPI_Comm comm_, MPI_Comm rom_com_
             }
             infile_offlineParam.close();
 
-            // Interpolate and save offset in the online phase of parametric save-and-load mode
+            // Compute and save the interplated offset
```
dylan-copeland: interpolated

rom/laghos_rom.cpp
```diff
@@ -711,6 +714,7 @@ void ROM_Basis::Init(ROM_Options const& input, Vector const& S)
 
     if ((offsetInit || spaceTime) && hyperreduce_prep)
     {
+        // Compute and save offset restricted on sample mesh in the online hyperreduction preparation phase for the all windows
```
dylan-copeland: `for the all` -> `for all`

dylan-copeland (approved):

PR #152, opened 2021-06-18
Sampling intermediate RK stages

siuwuncheung:
Implement the capability of sampling intermediate RK stages during offline stage. 

- Command line option `-sample-stages` controls the activation of the new capability.
- New member functions `GetRKStages` and `GetRKTime` are defined in the derived class `HydroODESolver` to enable the sampler access the RK stages and time stored in the time integrator.
- In order to enable the capability, `ode_solver` has to belong to a derived class of `HydroODESolver`, e.g. `RK2AvgSolver` and `RK4ROMSolver`. The new capability is now only compatible with RK2 average and RK4 schemes.

### DIFF
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -1504,8 +1504,8 @@ int main(int argc, char *argv[])
                 samplerTimer.Start();
                 if (rom_sample_stages)
                 {
-                    std::vector<Vector> RKStages = ode_solver_samp->GetRKStages();
-                    std::vector<double> RKTime = ode_solver_samp->GetRKTime();
+                    std::vector<Vector>& RKStages = ode_solver_samp->GetRKStages();
+                    std::vector<double>& RKTime = ode_solver_samp->GetRKTime();
                     MFEM_VERIFY(RKStages.size() == RKStepNumSamples, "Inconsistent number of Runge Kutta stages.");
                     for (int RKidx = 0; RKidx < RKStepNumSamples; ++RKidx)
                     {
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -667,33 +667,33 @@ int main(int argc, char *argv[])
     switch (ode_solver_type)
     {
     case 1:
-        ode_solver = new ForwardEulerSolver;
         rom_sample_stages = false;
+        ode_solver = new ForwardEulerSolver;
         if (rom_build_database) ode_solver_dat = new ForwardEulerSolver;
         break;
     case 2:
-        ode_solver = new RK2Solver(0.5);
         rom_sample_stages = false;
+        ode_solver = new RK2Solver(0.5);
         if (rom_build_database) ode_solver_dat = new RK2Solver(0.5);
         break;
     case 3:
-        ode_solver = new RK3SSPSolver;
         rom_sample_stages = false;
+        ode_solver = new RK3SSPSolver;
         if (rom_build_database) ode_solver_dat = new RK3SSPSolver;
         break;
     case 4:
-        ode_solver = new RK4ROMSolver(rom_online);
         if (rom_sample_stages) RKStepNumSamples = 3;
+        ode_solver = new RK4ROMSolver(rom_online);
         if (rom_build_database) ode_solver_dat = new RK4ROMSolver(rom_online);
         break;
     case 6:
-        ode_solver = new RK6Solver;
         rom_sample_stages = false;
+        ode_solver = new RK6Solver;
         if (rom_build_database) ode_solver_dat = new RK6Solver;
         break;
     case 7:
-        ode_solver = new RK2AvgSolver(rom_online, H1FESpace, L2FESpace);
         if (rom_sample_stages) RKStepNumSamples = 1;
+        ode_solver = new RK2AvgSolver(rom_online, H1FESpace, L2FESpace);
         if (rom_build_database) ode_solver_dat = new RK2AvgSolver(rom_online, H1FESpace, L2FESpace);
         break;
     default:
@@ -1481,6 +1481,20 @@ int main(int argc, char *argv[])
             }
 
             unique_steps++;
+
+            if (rom_sample_stages) 
+            {
+                // TODO: time this?
+                std::vector<Vector> RKStages = ode_solver_samp->GetRKStages();
+                std::vector<double> RKTime = ode_solver_samp->GetRKTime();
+                MFEM_VERIFY(RKStages.size() == RKStepNumSamples, "Inconsistent number of RK stages.");
+                for (int RKidx = 0; RKidx < RKStepNumSamples; ++RKidx)
+                {
+                    sampler->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
+                    if (samplerLast) samplerLast->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
+                    if (mpi.Root()) cout << "Runge Kutta stage " << RKidx+1 << " sampled" << endl;
+                }
+            }
 
             if (outputTimes) outfile_time << t << "\n";
 
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -1521,7 +1521,7 @@ int main(int argc, char *argv[])
                     }
                     else
                     {
-                        endWindow = (sampler->MaxNumSamples() + RKStepNumSamples >= windowNumSamples); 
+                        endWindow = (sampler->MaxNumSamples() + RKStepNumSamples * rom_sample_stages >= windowNumSamples); 
                     }
                 }
 
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -1481,20 +1481,6 @@ int main(int argc, char *argv[])
 
             unique_steps++;
 
-            if (rom_sample_stages) 
-            {
-                // TODO: time this?
-                std::vector<Vector> RKStages = ode_solver_samp->GetRKStages();
-                std::vector<double> RKTime = ode_solver_samp->GetRKTime();
-                MFEM_VERIFY(RKStages.size() == RKStepNumSamples, "Inconsistent number of Runge Kutta stages.");
-                for (int RKidx = 0; RKidx < RKStepNumSamples; ++RKidx)
-                {
-                    sampler->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
-                    if (samplerLast) samplerLast->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
-                    if (mpi.Root()) cout << "Runge-Kutta stage " << RKidx+1 << " sampled" << endl;
-                }
-            }
-
             if (outputTimes) outfile_time << t << "\n";
 
             if (outputSpaceTimeSolution)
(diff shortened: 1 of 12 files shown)


rom/laghos_rom.cpp
```diff
@@ -1285,8 +1285,8 @@ void ROM_Basis::SetupHyperreduction(ParFiniteElementSpace *H1FESpace, ParFiniteE
 
     // This creates sample_pmesh, sp_H1_space, and sp_L2_space only on rank 0.
     CAROM::CreateSampleMesh(*pmesh, H1_space, *H1FESpace, *L2FESpace, *(H1FESpace->FEColl()),
-                     *(L2FESpace->FEColl()), rom_com, sample_dofs_merged,
-                     num_sample_dofs_per_proc_merged, sample_pmesh, sprows, all_sprows, s2sp, st2sp, sp_H1_space, sp_L2_space);
+                            *(L2FESpace->FEColl()), rom_com, sample_dofs_merged,
```
dylan-copeland, line 1286: I wonder why `make style` did not this fix this already.

rom/laghos_timeinteg.cpp
```diff
@@ -42,6 +42,24 @@ void HydroODESolver::Init(TimeDependentOperator &_f)
     }
 }
 
+void HydroODESolver::SetSampler(ROM_Sampler *f)
+{
+    if (!rom)
```
dylan-copeland: This function should be called only if `!rom`, meaning if not online. Would it be better to do `MFEM_VERIFY(rom)`?

dylan-copeland:
@siuwuncheung After `laghos.cpp:1537` which has `samplerLast = NULL;`, should there be `if (rom_sample_stages) ode_solver_samp->SetSamplerLast(samplerLast);` to set the last sampler to `NULL` in the ODE solver?

rom/laghos.cpp
```diff
@@ -1494,7 +1504,7 @@ int main(int argc, char *argv[])
                     }
                     else
                     {
-                        endWindow = (sampler->MaxNumSamples() >= windowNumSamples);
+                        endWindow = (sampler->MaxNumSamples() >= windowNumSamples); // TODO: think about rom_sample_stages
```
dylan-copeland: Is there still an issue here?

dylan-copeland:
There are some failing regression tests, e.g. 
Verification failed: (rom_oper) is false:
 --> HydroSolvers expect ROM_Operator.
Also, some space-time regression tests fail, which I don't think is caused by this PR. The problem for those tests is 
Comparing: ROMsol/romS_
ROMsol exists on the baseline branch, but not on the user branch.
triple-point-space-time-romhr: FAIL
Maybe @kevinhkhuynh can help with the `ROMsol` issue.

rom/laghos_rom.cpp
siuwuncheung: This was indeed fixed by `make style`. Probably it was just we hadn't do `make style` before.

rom/laghos_utils.cpp
```diff
@@ -354,6 +355,8 @@ void SetWindowParameters(Array2D<int> const& twparam, ROM_Options & romOptions)
     romOptions.dimE = min(romOptions.max_dimE, twparam(w,2));
     romOptions.dimFv = romOptions.SNS ? romOptions.dimV : min(romOptions.max_dimFv, twparam(w,3));
     romOptions.dimFe = romOptions.SNS ? romOptions.dimE : min(romOptions.max_dimFe, twparam(w,4));
+    if (romOptions.useXV) romOptions.dimX = romOptions.dimV;
```
dylan-copeland, line 358: The `useXV` and `useVX` options have never been used, and SNS makes them even more obsolete. We probably should remove these options, not necessarily in this PR.

chldkdtn (approved):

rom/laghos_timeinteg.cpp
siuwuncheung: This routine has been removed as it is no longed needed.

chldkdtn:
@dylan-copeland and @kevinhkhuynh , Let's review and approve this PR today if possible.

dylan-copeland:
@siuwuncheung do the regression tests pass now?

kevinhkhuynh:
> 
> 
> There are some failing regression tests, e.g.
> Verification failed: (rom_oper) is false:
> --> HydroSolvers expect ROM_Operator.
> Also, some space-time regression tests fail, which I don't think is caused by this PR. The problem for those tests is
> Comparing: ROMsol/romS_
> ROMsol exists on the baseline branch, but not on the user branch.
> triple-point-space-time-romhr: FAIL
> Maybe @kevinhkhuynh can help with the `ROMsol` issue.

Are the space-time romhr regression tests working? I have run them manually and no romS files are outputted. Is that intended? If so, then I'll need to update the test script.

siuwuncheung:
@dylan-copeland As @kevinhkhuynh has said, the space-time tests are failing. The `ROMsol` directory for both baseline and user branch in these tests are empty.

rom/laghos.cpp
```diff
@@ -1494,6 +1525,7 @@ int main(int argc, char *argv[])
                     }
                     else
                     {
+                        //endWindow = (sampler->MaxNumSamples() + RKStepNumSamples * rom_sample_stages >= windowNumSamples);
```
dylan-copeland: Can this be removed?

rom/merge.cpp
```diff
@@ -306,6 +306,8 @@ void GetParametricTimeWindows(const int nset, const bool SNS, const std::string&
         double overlapMidpoint = (windowLeft + windowRight) / 2;
         twep.Append(overlapMidpoint);
 
+        cout << windowLeft << " " << windowRight << endl;
```
dylan-copeland: Is this just for debugging, or is it generally useful? If it should be kept, is it clear what the output is (no label in the output)?

rom/laghos_timeinteg.hpp
```diff
@@ -45,6 +48,16 @@ class HydroODESolver : public ODESolver
     {
         MFEM_ABORT("Time stepping is undefined.");
     }
+
+    std::vector<Vector> GetRKStages()
```
dylan-copeland: It seems the return value should be by reference, to avoid copying the data.

rom/laghos_timeinteg.cpp
```diff
@@ -50,6 +50,8 @@ void RK2AvgSolver::Step(Vector &S, double &t, double &dt)
         return;
     }
 
+    RKStages.clear();
```
dylan-copeland: It is unclear whether this would have a significant performance impact, but clearing the vectors and pushing back vectors on each timestep is less efficient than keeping memory allocated throughout the simulation and copying. That is, the constructor for the solver can allocate as many stages as needed in `RKStages`, and then on each stage just copy `S` to `RKStages[i]`.

rom/laghos_rom.cpp
```diff
@@ -165,6 +165,10 @@ void printSnapshotTime(std::vector<double> const &tSnap, std::string const path,
 {
     cout << var << " snapshot size: " << tSnap.size() << endl;
     std::ofstream outfile_tSnap(path + var);
+
+    typedef std::numeric_limits< double > dbl;
```
dylan-copeland: Instead of the `typedef` could there just be `std::numeric_limits<double>::max_digits10`?

rom/laghos_rom.cpp
```diff
@@ -1775,6 +1774,7 @@ void ROM_Basis::RestrictFromSampleMesh(const Vector &usp, Vector &u, const bool
     for (int i=0; i<numSamplesE; ++i)
         (*sE)(i) = useOffset ? usp[(2*size_H1_sp) + s2sp_E[i]] - (*initEsp)(s2sp_E[i]) : usp[(2*size_H1_sp) + s2sp_E[i]];
 
+    // ROM opertion on source: map sample mesh evaluation to reduced coefficients with respect to solution bases
```
dylan-copeland: opertion -> operation

dylan-copeland:
I confirmed that only the space-time regression tests are failing, and the reason is that the `romS_*` files are not written. These are ROM solutions written to file on each timestep in the spatial ROM case. For space-time, there is only one solution for all time, not solutions at each time. There is no `restore` phase for space-time, yet. When that is implemented, these files at each timestep could be written out to files for comparison.

I am not sure why this fails now but did not before. One solution would be to skip the comparison of these files in the space-time case, but that may be complicated for the regression testing scripts. @kevinhkhuynh do you think this would be easy? We can (i) let the tests fail for now and fix this later, (ii) skip this check for space-time tests (@kevinhkhuynh would need to do that), or (iii) implement a restore phase. For this PR, (i) seems most appropriate, unless (ii) can be done quickly.

chldkdtn:
> I confirmed that only the space-time regression tests are failing, and the reason is that the `romS_*` files are not written. These are ROM solutions written to file on each timestep in the spatial ROM case. For space-time, there is only one solution for all time, not solutions at each time. There is no `restore` phase for space-time, yet. When that is implemented, these files at each timestep could be written out to files for comparison.
> 
> I am not sure why this fails now but did not before. One solution would be to skip the comparison of these files in the space-time case, but that may be complicated for the regression testing scripts. @kevinhkhuynh do you think this would be easy? We can (i) let the tests fail for now and fix this later, (ii) skip this check for space-time tests (@kevinhkhuynh would need to do that), or (iii) implement a restore phase. For this PR, (i) seems most appropriate, unless (ii) can be done quickly.

@dylan-copeland I am okay if we take Option (ii).

kevinhkhuynh:
> 
> 
> I confirmed that only the space-time regression tests are failing, and the reason is that the `romS_*` files are not written. These are ROM solutions written to file on each timestep in the spatial ROM case. For space-time, there is only one solution for all time, not solutions at each time. There is no `restore` phase for space-time, yet. When that is implemented, these files at each timestep could be written out to files for comparison.
> 
> I am not sure why this fails now but did not before. One solution would be to skip the comparison of these files in the space-time case, but that may be complicated for the regression testing scripts. @kevinhkhuynh do you think this would be easy? We can (i) let the tests fail for now and fix this later, (ii) skip this check for space-time tests (@kevinhkhuynh would need to do that), or (iii) implement a restore phase. For this PR, (i) seems most appropriate, unless (ii) can be done quickly.

I just moved the online phase into the offline for now, meaning the online phase still runs, but only the offline stuff is checked.

kevinhkhuynh (approved):

rom/laghos_rom.cpp
siuwuncheung: Yes it can and should be changed. There is a point when the code does not work and I add this `typedef`, but I later figure out that is not the source of the problem.

rom/merge.cpp
siuwuncheung: This is just for debugging. I should remove it.

rom/laghos.cpp
siuwuncheung: Yes it can be safely removed.

rom/laghos_rom.cpp
siuwuncheung: Fixed.

rom/laghos.cpp
```diff
@@ -1509,7 +1509,7 @@ int main(int argc, char *argv[])
                     MFEM_VERIFY(RKStages.size() == RKStepNumSamples, "Inconsistent number of Runge Kutta stages.");
                     for (int RKidx = 0; RKidx < RKStepNumSamples; ++RKidx)
                     {
-                        sampler->SampleSolution(RKTime[RKidx], dt, RKStages[RKidx]);
+                        sampler->SampleSolution(RKTime[RKidx], last_dt, RKStages[RKidx]);
```
dylan-copeland, line 1512: I'm not sure how the sampler works in libROM (would need to review the code), but I wonder whether it matters to the sampler and its determination of when to sample that multiple samples are taken within the same timestep, with the same timestep size. That is, this is inconsistent input to the sampler. I don't think it matters for Laghos, as we end up sampling every timestep anyway, so we don't actually rely on the sampler to accurately determine when to sample. I am not saying this needs to be changed, just commenting that it is strange and perhaps should be thought about more in the future.

siuwuncheung: Yes, I think we also have take a similar shortcut by setting an artificial time and timestep in `mergeXV`. Though the code works and every snapshots are sampled, we should review libROM and avoid the potential misuse of `dt`.

chldkdtn:
@siuwuncheung @dylan-copeland is this PR ready to merge?

rom/laghos_timeinteg.cpp
siuwuncheung: Resolved.

dylan-copeland (approved):

siuwuncheung:
> @siuwuncheung @dylan-copeland is this PR ready to merge?

Now the PR gets all the approval and is ready to merge.

PR #155, opened 2021-08-30
User-specified set of samples

siuwuncheung:
This PR, used along with [libROM PR#72](https://github.com/LLNL/libROM/pull/72), allows specifying an arbitrary number `MV, ME` of samples to be used in hyperreduction. It assumes a csv file format with the filenames specified by `initSamples_basename + "V.csv"` and `initSamples_basename + "E.csv"` placed in `outputPath`, where each row specifies a sample. The actual total number of samples `NV, NE` used is bounded by the options `-sfacv, -sface` or `-nsamv, -nsame`. Several cases in the usage (`*` stands for `V` or `E`): 

- `initSamples_basename + "*.csv"` is absent: use the greedy algorithm to sample the `N*` indices as usual
- `initSamples_basename + "*.csv"` is present: read all the `M*` specified samples in the file. If `N* > M*`, use all the `M*` specified samples and use the greedy algorithm to sample the remaining`N*-M*` indices. Otherwise, use only the first `N*` specified samples and no greedy algorithm will be casted.


rom/laghos_rom.cpp
```diff
@@ -951,6 +951,42 @@ void ROM_Basis::SetupHyperreduction(ParFiniteElementSpace *H1FESpace, ParFiniteE
         cout << "number of samples for energy  : " << numSamplesE << "\n";
     }
 
+    int numInitSamplesV = 0; 
+    initSamplesV.clear();
+    std::string initSamplesV_filename = basename + "/initSamplesV.csv"; // TODO: change to hyperreduce_basename
```
dylan-copeland: Should the name changed?

dylan-copeland (approved):
I ran an experiment with initial samples and confirmed it works (with one randomly chosen sample). If many samples are chosen randomly, then the online simulation may fail due to bad sampling, which makes sense, so prescribing samples must be done in a sensible way, of course. The only issue I see with this PR is that it is not very clear or user-friendly to define the CSV files for the initial samples. The only way to know about this is to read the code in laghos_rom.cpp, which is hard to find. In other words, a typical user will not even know this option exists, as it is not documented. Also, to know where to put the CSV files, the user needs to know what directory `basename` is. For our own internal use, this is fine for now, but depending on how widely this is used, we may want to improve the interface.

rom/laghos_rom.cpp
```diff
@@ -951,35 +952,38 @@ void ROM_Basis::SetupHyperreduction(ParFiniteElementSpace *H1FESpace, ParFiniteE
...
     }
 
-    int numInitSamplesV = 0; 
+    // Read the initial samples from file
+    // TODO: window-dependent initialization is not supported yet.
+
+    int numInitSamplesV = 0;
     initSamplesV.clear();
-    std::string initSamplesV_filename = basename + "/initSamplesV.csv"; // TODO: change to hyperreduce_basename
+    std::string initSamplesV_filename = basename + "/" + initSamples_basename + "V.csv"; // TODO: change to hyperreduce_basename
```
dylan-copeland: "TODO" can be removed.

siuwuncheung: Updated.

rom/laghos_rom.cpp
siuwuncheung: Updated.

rom/laghos.cpp
```diff
@@ -218,6 +219,8 @@ int main(int argc, char *argv[])
                    "Name of the CSV file defining offline time windows");
     args.AddOption(&twpfile, "-twp", "--timewindowparamfilename",
                    "Name of the CSV file defining online time window parameters");
+    args.AddOption(&initSamples_basename, "-is", "--initsamplesfileename",
```
kevinhkhuynh: typo initsamplesfilename

kevinhkhuynh (approved):

chldkdtn (approved):

rom/laghos.cpp
siuwuncheung: Resolved.

PR #160, opened 2021-11-04
Partitioning bug fix

dylan-copeland:
Fixing a bug and some typos. The regression tests all pass with these changes.

### DIFF
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -523,7 +523,7 @@ int main(int argc, char *argv[])
     if (fom_data)
     {
         const int num_tasks = mpi.WorldSize();
-        int unit;
+        int unit = 1;
         int *nxyz = new int[dim];
         switch (partition_type)
         {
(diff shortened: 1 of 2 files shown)


rom/laghos.cpp
```diff
@@ -523,7 +523,7 @@ int main(int argc, char *argv[])
     if (fom_data)
     {
         const int num_tasks = mpi.WorldSize();
-        int unit;
+        int unit = 1;
```
chldkdtn, line 526: Wow, it was not initialized before? I wonder how this was even working in the past?

chldkdtn (approved):

rom/laghos.cpp
dylan-copeland: This is a special case (`partition_type==0`) that is apparently not used in any of the parallel regression tests.

kevinhkhuynh (approved):

PR #161, opened 2021-11-04
Shifted energy factor calculation

siuwuncheung:
Shift energy factor calculation for time windowing when initial offset is used.


dylan-copeland:
@siuwuncheung it makes sense to add the option to shift the singular values, and my only question is whether it is generally good to hard-code the shifting as done in this PR. What is the reason that shifting by 1 is always good when using initial offsets, for any problem? If there is a good reason, it should be documented. Or would it make sense to have an input parameter to set the shift?

siuwuncheung:
> @siuwuncheung it makes sense to add the option to shift the singular values, and my only question is whether it is generally good to hard-code the shifting as done in this PR. What is the reason that shifting by 1 is always good when using initial offsets, for any problem? If there is a good reason, it should be documented. Or would it make sense to have an input parameter to set the shift?

Theoretically shifting the singular value will give more accurate results but less speed-up with a larger dimension. But based on the experience, the numerical instability problem happens to insufficient basis dimension and the time step size eventually gets small (complete with no speed-up) or even vanishes (never complete). At the same energy factor level, the initial offset option gives less basis vectors than the interpolated offset, it does not actually achieve a better overall speed-up than the interpolated offset in most of the experiments (except the long-time Rayleigh-Taylor where interpolation does not work well). When I checked the basis number, sometimes the position basis just has 1 vector since itself occupies 99.99% energy factor and the original truncation is too aggressive), making the initial offset scheme useless. (We can always make energy factor larger to add more, but I believe it is better to have an “automatic” way to avoid the case of only 1 vector.) On the other hand, the offset in the interpolation option (or the deprecated previous state option) works like a dominant component, makes the decay of singular values more moderate. To make a fair comparison, I think we should not count the first singular value when the initial offset is used. I think it is fair to give up a bit of “theoretical speed-up” for a more reliable scheme using the initial offset, and perhaps we will see more scenarios where it is more useful than interpolation. If we want to leave the freedom of not shifting to compare the results, I can add an input parameter to set the shift.

dylan-copeland:
@siuwuncheung thanks for the explanation, and I agree it is a good improvement. I think it would be better to make this a user input parameter, with a default value of 1, so that only an expert user would change it (maybe just to verify it is the best option). Also, some documentation about these issues would be helpful.

rom/laghos_rom.hpp
```diff
@@ -383,6 +384,7 @@ class ROM_Sampler
                 }
             }
         }
+        else first_sv = 1;
```
dylan-copeland: Isn't this the case of no offsets? Does the first SV still need to be skipped in that case?

rom/merge.cpp
```diff
@@ -146,14 +146,15 @@ void LoadSampleSets(const int rank, const double energyFraction, const int nsets
     }
     std::string basis_filename = basename + "/basis" + varName + std::to_string(basisWindow) + basisIdentifier;
 
+    int first_sv = (useOffset && offsetType == useInitialState && basisWindow > 0) && (v == X || v == V || v == E);;
```
dylan-copeland: extra `;`

chldkdtn:
> @siuwuncheung thanks for the explanation, and I agree it is a good improvement. I think it would be better to make this a user input parameter, with a default value of 1, so that only an expert user would change it (maybe just to verify it is the best option). Also, some documentation about these issues would be helpful.

Making this a user input parameter with a default value of 1 sounds great to me.

chldkdtn:
@siuwuncheung are all the things commented above addressed?

rom/laghos_rom.hpp
siuwuncheung: Yes this is the case of no offsets, which is rarely used. However, if it is used, it has the same issue: there is one single principal component which eats up too much energy factor (even in shorter time simulation in this case). So I do the same as in the initial offset.

rom/merge.cpp
siuwuncheung: Thanks for pointing that out! I just removed the extra one.

siuwuncheung:
> @siuwuncheung are all the things commented above addressed?

A user-defined option is added. I am setting the default value to be false (no shift), and performing the regression tests.  All the regression tests are expected to pass. Once verified, we can turn the default value to be true.

siuwuncheung:
Regression tests all passed in commit 08553333f8d349e669f9066e4bbb8528e0bbe83c. If everyone agrees, this branch can be merged after setting the default value of `shift_sv` to be true.

chldkdtn (approved):

rom/laghos.cpp
```diff
@@ -264,6 +264,8 @@ int main(int argc, char *argv[])
                    "Energy fraction for recommended ROM basis sizes.");
     args.AddOption(&romOptions.energyFraction_X, "-efx", "--rom-efx",
                    "Energy fraction for recommended X ROM basis size.");
+    args.AddOption(&romOptions.shift_sv, "-shiftsv", "--shift-svv", "-no-shiftsv", "--no-shiftsv",
```
dylan-copeland: `--shift-svv` looks like a typo.

rom/laghos_rom.hpp
```diff
@@ -169,6 +169,7 @@ struct ROM_Options
     bool SNS = false; // if true, use SNS relation to obtain nonlinear RHS bases by multiplying mass matrix to a solution matrix. See arXiv 1809.04064.
     double energyFraction = 0.9999; // used for recommending basis sizes, depending on singular values
     double energyFraction_X = 0.9999; // used for recommending basis sizes, depending on singular values
+    bool shift_sv = false; // if true, shift energy fraction calculation when window-dependent offsets are not used
```
dylan-copeland: Should this be an `int` type, so that any shift could be specified (default 1)? Also, it would be helpful to have a more detailed comment, perhaps here, explaining why the shift is needed.

rom/merge.cpp
```diff
@@ -408,6 +416,8 @@ int main(int argc, char *argv[])
     args.AddOption(&windowOverlapSamples, "-nwinover", "--numwindowoverlap", "Number of samples for ROM window overlap.");
     args.AddOption(&basisIdentifier, "-bi", "--bi", "Basis identifier for the greedy algorithm.");
     args.AddOption(&romOptions.energyFraction, "-ef", "--rom-ef", "Energy fraction for recommended ROM basis sizes.");
+    args.AddOption(&romOptions.shift_sv, "-shiftsv", "--shift-svv", "-no-shiftsv", "--no-shiftsv",
```
dylan-copeland: `--shift-svv` again (see other comment)

dylan-copeland (changes requested):

dylan-copeland (approved):

kevinhkhuynh (approved):

PR #173, opened 2023-03-07
EQP

dylan-copeland:


chldkdtn (approved):

chldkdtn:
Is it ready for review and merge?

dylan-copeland:
> Is it ready for review and merge?

@chldkdtn it is working in some cases, but I am planning to test more simulations and possibly fix some issues for parallel tests.

chldkdtn:
@dylan-copeland Let's keep in mind that we need regression tests for EQP.

rom/laghos_rom.cpp
```diff
@@ -1893,6 +2489,101 @@ ROM_Operator::ROM_Operator(ROM_Options const& input, ROM_Basis *b,
         ComputeReducedMv();
         ComputeReducedMe();
     }
+
+    if (hyperreduce && hyperreductionSamplingType == eqp)
+    {
+        // TODO: are reduced mass matrices needed for EQP, or just W matrices?
```
cval26: For EQP, I believe that we do not use reduced mass matrices. Nevertheless, it is the `ComputeReducedMv()` function that computes the W matrices needed for forming the velocity quadrature rule (similarly for the energy).

rom/laghos_rom.cpp
```diff
@@ -2197,7 +2887,7 @@ void ROM_Operator::ComputeReducedMv()
 {
     const int nv = basis->GetDimV();
 
-    if (hyperreduce && rank == 0)
+    if (use_sample_mesh && rank == 0)
```
cval26: I'm a bit confused about the logic of this if block. If we're using EQP hyperreduction and are in the hyperreduction phase, then both `hyperreduce` (and thus `use_sample_mesh`) and `hyperreductionSamplingtype == eqp` yield true conditions, so both branches of the if block are true, but only one will be accessed (the first one). 

Is that by design? And if so, where are the required W matrices computed then?

The same question applies to `ComputeReducedMe()` below.

EDIT: I think that the W matrices are computed in the offline stage and written to files; so in the online stage we only to read them, not compute them.

rom/laghos.cpp
```diff
@@ -395,10 +413,12 @@ int main(int argc, char *argv[])
     romOptions.basisIdentifier = std::string(basisIdentifier);
 
     romOptions.hyperreductionSamplingType = getHyperreductionSamplingType(hyperreductionSamplingType);
+    romOptions.use_sample_mesh = romOptions.hyperreduce;
```
cval26: So the option `use_sample_mesh` is true for all hyperreduction types?

rom/laghos_eqp.cpp
```diff
@@ -0,0 +1,1102 @@
...
+                W(i,j) = g_i[i];
+#endif
+        }
+        else
+        {
+            for (int i=0; i<tH1size; ++i)
+                v_i[i] = (*basisV)(i,j);
+
+            gfH1->SetFromTrueDofs(v_i);
+#ifdef EQP_NO_Minv
```
cval26, line 563: Is there a runtime flag we can use to check that, instead of using the preprocessor?

rom/laghos_rom.cpp
dylan-copeland: Good catch. This doesn't make sense, and it resulted from having different versions that changed over time. In the initial working implementation of EQP, there was no sample mesh, and the FOM mesh and spaces were maintained. The speedup came from integrating only on sampled elements, although all FOM elements were available. In that early version, EQP had `use_sample_mesh = false` and `hyperreduce = true`. Now that EQP uses a sample mesh for better performance, this distinction is no longer necessary. I removed `use_sample_mesh` in cd6c9a1bdfdb8912fc669b250cee1684caf63fac, and the code is less confusing now.

rom/laghos_rom.cpp
dylan-copeland: Good point, I was able to remove the calls to `ComputeReducedMv` and `ComputeReducedMe` in cd6c9a1bdfdb8912fc669b250cee1684caf63fac and everything still works.

rom/laghos.cpp
dylan-copeland: Yes, this option is removed now.

rom/laghos_eqp.cpp
dylan-copeland: Usually we select options with runtime flags, and that could be done here. In this case, the option to not use mass matrix inverses at all gives very bad results, and there are no cases in which this is known to be a good option. Some possibilities are

1. Remove the code in `#ifdef EQP_NO_Minv`, since it should not be used.
2. Add a runtime flag to choose whether to use mass matrix inverses.
3. Keep the code, so anyone curious can easily verify that it is a bad option.

I chose 3, just for easy testing in the future, without indicating that it is intended as a normal option.

rom/laghos_rom.cpp
cval26: Understood, thanks for clarifying. So the computation of the W matrices does indeed happen during the offline stage?

For what it's worth, I think the if block in the updated `ComputeReducedMv()` function is still ambiguous; in the case that we're using `hyperreduce` and `hyperreductionSamplingType==EQP` both of the first two clauses should be accessed, but only the first one will be.

Now that you removed calls to this function this might not make a difference, but I wonder if it's affecting other areas of the code.

rom/laghos.cpp
```diff
@@ -134,6 +134,7 @@ int main(int argc, char *argv[])
     double blast_energy = 0.25;
     double blast_position[] = {0.0, 0.0, 0.0};
     double dt_factor = 1.0;
+    double windowTol = 0.99;
```
siuwuncheung: Should we use the default value of 1? I am wondering if this will pass regression tests.

rom/laghos.cpp
```diff
@@ -163,6 +164,8 @@ int main(int argc, char *argv[])
     Array<double> twep;
     Array2D<int> twparam;
     ROM_Options romOptions;
+    int nnlsWindow0 = -1;
```
siuwuncheung, line 167: Maybe we need a documentation for these new ideas, e.g., in the Overleaf document.

rom/laghos.cpp
```diff
@@ -1310,7 +1349,7 @@ int main(int argc, char *argv[])
                 for (int curr_window = numWindows-1; curr_window >= 0; --curr_window)
                     basis[curr_window]->writePDweights(pd2_tdof, curr_window);
             }
-            if (!romOptions.hyperreduce_prep)
+            if (romOptions.hyperreduce)
```
siuwuncheung: If I remember correctly, we intentionally leave the option of online ROM via Galerkin projection but **NOT** using hyper-reduction, which I believe is because we want to see the fact that hyper-reduction is needed for speed-up. But we ended up having these strange conditions `!romOptions.hyperreduce_prep`. Should we get rid of this consideration from now on?

rom/laghos.cpp
```diff
@@ -2082,6 +2131,22 @@ int main(int argc, char *argv[])
                              << sqrt(tot_norm) << endl;
                     }
 
+                    if (romOptions.hyperreductionSamplingType == eqp)
```
siuwuncheung: Actually, it will be interesting to be able to see how the energy conservation is like for any kind of hyper-reduction instead of limiting it to EQP. How about we make an option like `calculate_energy`?

rom/laghos_rom.cpp
dylan-copeland: The `W` matrices are written in the `romhrprep` stage (online preparation), by `ROM_Basis::SetupEQP_Force`. They are read in the online stage `romhr`, by the `ROM_Operator` constructor. You are right that the code in `ComputeReducedMv` to compute `W` is not used. Instead, these matrices are computed in `ROM_Basis::SetupEQP_Force`. I removed the unused code in 3581ea7626134f3c99a608a8e3ed143cf9dc8309

rom/laghos.cpp
dylan-copeland: Right, I had a note to change this to 1 but forgot. It is changed in 3581ea7626134f3c99a608a8e3ed143cf9dc8309

rom/laghos.cpp
dylan-copeland: This change looks like a mistake. It has nothing to do with this PR. I will revert this change, see b648ac15186e1163d6761297699fcde4a7575f62. You still have a valid question, which I think should be addressed outside this PR.

rom/laghos.cpp
dylan-copeland: It turns out this code is not reached, so I will remove it, see b648ac15186e1163d6761297699fcde4a7575f62. The energy change is output at the end, around line 2523. This PR added more details about the energy change there.

rom/laghos.cpp
dylan-copeland: I agree that some high-level documentation outside the code would be very helpful. Of course, more comments in the code are always needed as well.

cval26 (approved):
Dylan, just to make sure the changes made during the review process did not break anything, I suggest that you run a representative simulation before merging, to check that you get the same accuracy results.

siuwuncheung (approved):

PR #177, opened 2023-06-28
Energy-conserving EQP

cval26:
I'm opening this draft PR so that you can stay up to date with my changes in the code, and so that I can ask you questions about specific parts of the code that I'm unsure of. I'll add comments to the parts of the code I have questions on. 

I'll push new changes as I implement them so that we can discuss them. Once we have a correct implementation I can go back and edit the commit history so that it's easier to review.

The indentation in some parts of the code seems to be off in the diff files. I'm not sure why; it doesn't look like that in my editor. 

Changes made so far: 

- formed a new, combined NNLS problem for both velocity and energy;
- formed the new RHS force vectors used in the online timestepping phase;
- added the energy identity vector to each time window's reduced energy basis;

### DIFF
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -345,6 +345,8 @@ int main(int argc, char *argv[])
                    "Maximum number of nonzeros in NNLS solution.");
     args.AddOption(&romOptions.tolNNLS, "-tolnnls", "--tol-nnls",
                    "NNLS solver error tolerance.");
+    args.AddOption(&romOptions.sampfreq, "-sampfreq", "--samp-freq",
+                   "Snapshot sampling frequency.");
     
 	args.Parse();
     if (!args.Good())
@@ -1178,6 +1180,10 @@ int main(int argc, char *argv[])
             sampler->SampleSolution(0, 0, (problem == 7) ? 0.0 : -1.0, *S);
         }
         samplerTimer.Stop();
+
+		if (myid == 0)
+			cout << "Sampling every " << romOptions.sampfreq <<
+				" timestep(s)." << endl;
     }
 
     if (outputTimes)
@@ -1820,7 +1826,8 @@ int main(int argc, char *argv[])
                 }
                 else
                 {
-                    sampler->SampleSolution(t, last_dt, real_pd, *S);
+					if (unique_steps % romOptions.sampfreq == 0) 
+						sampler->SampleSolution(t, last_dt, real_pd, *S);
                 }
 
                 bool endWindow = false;
@@ -1986,6 +1993,22 @@ int main(int argc, char *argv[])
 
             if (rom_online)
             {
+				if (romOptions.hyperreductionSamplingType == eqp_energy)
+				{
+					double energy_total, energy_diff;
+					energy_total = oper->InternalEnergy(*e_gf) +
+						oper->KineticEnergy(*v_gf);
+					energy_diff	= energy_total - energy_init;
+
+					if (mpi.Root())
+					{
+						cout << "\tE_tot = " << scientific << setprecision(5)
+							<< energy_total
+							<< ",\tE_diff = " << scientific << setprecision(5)
+							<< energy_diff << endl; 
+					}
+				}
+
                 double window_par = t;
                 if (problem == 7)
                 {
@@ -2049,10 +2072,31 @@ int main(int argc, char *argv[])
                     {
                         basis[romOptions.window]->ProjectFOMtoROM(*S, romS);
                     }
-                    if (myid == 0)
-                    {
-                        cout << "Window " << romOptions.window << ": initial romS norm " << romS.Norml2() << endl;
-                    }
+					if (myid == 0)
+					{
+						cout << "Window " << romOptions.window << ": initial romS norm " << romS.Norml2() << endl;
+					}
+
+					// Recompute and print the energy information after the
+					// change of basis, before any more timestepping takes
+					// place.
+					if (romOptions.hyperreductionSamplingType == eqp_energy)
+					{
+						basis[romOptions.window]->LiftROMtoFOM(romS, *S);
+
+						double energy_total, energy_diff;
+						energy_total = oper->InternalEnergy(*e_gf) +
+							oper->KineticEnergy(*v_gf);
+						energy_diff	= energy_total - energy_init;
+
+						if (mpi.Root())
+						{
+							cout << "\tE_tot = " << scientific << setprecision(5)
+								<< energy_total
+								<< ",\tE_diff = " << scientific << setprecision(5)
+								<< energy_diff << endl; 
+						}
+					}
 
                     delete romOper[romOptions.window-1];
 
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -395,13 +395,13 @@ int main(int argc, char *argv[])
 
     romOptions.hyperreductionSamplingType = getHyperreductionSamplingType(hyperreductionSamplingType);
     romOptions.use_sample_mesh = romOptions.hyperreduce && (romOptions.hyperreductionSamplingType != eqp
-			|| romOptions.hyperreductionSamplingType != eqp_energy);
+			&& romOptions.hyperreductionSamplingType != eqp_energy);
 
     MFEM_VERIFY(!(romOptions.SNS) || (romOptions.hyperreductionSamplingType != eqp &&
 			romOptions.hyperreductionSamplingType != eqp_energy),
 			"Using SNS with EQP is prohibited");
 	
-	if (romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
+	if (romOptions.hyperreductionSamplingType == eqp_energy)
 		romOptions.GramSchmidt = false;
 
     romOptions.spaceTimeMethod = getSpaceTimeMethod(spaceTimeMethod);
@@ -1004,14 +1004,14 @@ int main(int argc, char *argv[])
     LagrangianHydroOperator* oper = NULL;
     if (fom_data)
     {
-        const bool noMassSolve = rom_online && (romOptions.hyperreductionSamplingType == eqp);
+        const bool noMassSolve = rom_online && (romOptions.hyperreductionSamplingType == eqp || romOptions.hyperreductionSamplingType == eqp_energy);
         oper = new LagrangianHydroOperator(S->Size(), *H1FESpace, *L2FESpace,
                                            ess_tdofs, *rho, source, cfl,
                                            mat_gf_coeff, visc, vort, p_assembly,
                                            cg_tol, cg_max_iter, ftz_tol,
                                            H1FEC.GetBasisType(), noMassSolve,
                                            noMassSolve,
-                                           rom_online && (romOptions.hyperreductionSamplingType == eqp));
+                                           rom_online && (romOptions.hyperreductionSamplingType == eqp || romOptions.hyperreductionSamplingType == eqp_energy));
     }
 
     socketstream* vis_rho = NULL;
@@ -1581,11 +1581,11 @@ int main(int argc, char *argv[])
             romOper[0]->ApplyHyperreduction(romS);
         }
 
-		// TODO: do we want that for the energy-conserving EQP?
-		if (rom_online && romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
-		{
-			romOper[0]->ApplyHyperreduction(romS);
-		}
+	//	// TODO: do we want that for the energy-conserving EQP?
+	//	if (rom_online && romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
+	//	{
+	//		romOper[0]->ApplyHyperreduction(romS);
+	//	}
 
 		double windowEndpoint = 0.0;
         double windowOverlapMidpoint = 0.0;
@@ -2058,10 +2058,10 @@ int main(int argc, char *argv[])
                         romOper[romOptions.window]->ApplyHyperreduction(romS);
                     }
 		
-					if (romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
-					{
-						romOper[romOptions.window]->ApplyHyperreduction(romS);
-					}
+					//if (romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
+					//{
+					//	romOper[romOptions.window]->ApplyHyperreduction(romS);
+					//}
 
                     if (problem == 7 && romOptions.indicatorType == penetrationDistance)
                     {
@@ -2109,7 +2109,7 @@ int main(int argc, char *argv[])
                              << sqrt(tot_norm) << endl;
                     }
 					
-					if (romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
+					if (romOptions.hyperreductionSamplingType == eqp_energy)
 					{
 						double energy_total, energy_diff;
 						energy_total = oper->InternalEnergy(*e_gf) +
--- a/rom/laghos.cpp
+++ b/rom/laghos.cpp
@@ -1914,7 +1914,7 @@ int main(int argc, char *argv[])
 
                         if (samplerLast->MaxNumSamples() >= windowNumSamples + windowOverlapSamples || last_step)
                         {
-                            samplerLast->Finalize(cutoff, romOptions);
+                            samplerLast->Finalize(cutoff, romOptions, *S);
                             if (last_step)
                             {
                                 // Let samplerLast define the final window, discarding the sampler window.
@@ -1960,7 +1960,7 @@ int main(int argc, char *argv[])
                         else
                         {
 							// Form the reduced bases from the snapshots.
-                            sampler->Finalize(cutoff, romOptions);
+                            sampler->Finalize(cutoff, romOptions, *S);
                         }
                         if (myid == 0 && romOptions.parameterID == -1) {
                             outfile_twp << windowEndpoint << ", " << cutoff[0] << ", " << cutoff[1] << ", " << cutoff[2];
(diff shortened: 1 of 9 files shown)


rom/laghos_rom.cpp
```diff
@@ -608,8 +971,17 @@ void ROM_Sampler::SetupEQP_Force(const CAROM::Matrix* snapX, const CAROM::Matrix
...
 
-    SetupEQP_Force_Eq(snapX, snapV, snapE, basisV, basisE, input, false);
-    SetupEQP_Force_Eq(snapX, snapV, snapE, basisV, basisE, input, true);
+	if (input.hyperreductionSamplingType == eqp)
+	{
+		// basic EQP: different rules for velocity and energy
+		SetupEQP_Force_Eq(snapX, snapV, snapE, basisV, basisE, input, false);
+		SetupEQP_Force_Eq(snapX, snapV, snapE, basisV, basisE, input, true);
+	}
+	else if (input.hyperreductionSamplingType == eqp_energy)
```
cval26, line 980: Where should I add the hyperreduction option `eqp_energy` so that it's a valid option? This is the only part of the code that gives me a compilation error.

rom/laghos_rom.cpp
```diff
@@ -338,6 +339,84 @@ void ComputeElementRowOfG_E(const IntegrationRule *ir,
...
+		loc_force *= 1.0 / ip.weight;  // Divide by exact quadrature weight
+
+		if (equationV)
+		{
+			// set rv = sum(Vloc_force) 
+			rv[q] = 0.0;
+			for (int i = 0; i < h1dofs_cnt * dim; i++)
+			{
+				rv[q] += Vloc_force(i);
+			}
```
cval26, line 405: For the velocity constraints I simply sum the `Vloc_force`, since the force vector for the velocity equation is just the integral of `stress : grad w_j`, with `w_j` a velocity basis vector. This is what's computed in the loop above.

rom/laghos_rom.cpp
```diff
@@ -338,6 +339,84 @@ void ComputeElementRowOfG_E(const IntegrationRule *ir,
...
+		if (equationE)
+		{
+			// set re = F^T * v; F = integral of (stress : grad w_i)phi_j
+			re[q] = 0.0;
+			for (int i = 0; i < l2dofs_cnt; i++)
+			{
+				re[q] += w_e(i);
+			}
+			re[q] *= v_e * Vloc_force;
+		}
```
cval26, line 412: For the energy equation I take the inner product of `Vloc_force` with the velocity snapshot vector `v_e` and multiply it with the sum of `w_e`, with `w_e` an energy basis vector. This is to form the product `F^T * v`, where

- `F` is integral of `(stress : grad w_j) phi_i`, with `w_j` a velocity basis vector and `phi_i` an energy basis vector;
- `v` is a velocity snapshot vector. 

I tried to keep the approach similar to that of the current EQP implementation, but I'm not sure if this yields the correct result.

rom/laghos_rom.cpp
```diff
@@ -597,9 +695,274 @@ void ROM_Sampler::SetupEQP_Force_Eq(const CAROM::Matrix* snapX,
...
+	CAROM::Matrix We(L2size, NBe, true);
+
+	// jth Wv columnn holds the jth velocity basis vector
+	// TODO: do we need Wv anymore? it's not used anywhere
+	for (int j=0; j<NBv; ++j)
+	{
+		for (int i=0; i<tH1size; ++i) v_i[i] = (*basisV)(i,j);
+		gfH1.SetFromTrueDofs(v_i);
+		for (int i=0; i<H1size; ++i) Wv(i,j) = gfH1[i];
+	}
```
cval26, line 757: Given the changes made in our choice for force vectors, I believe that `Wv` is not used anymore. It was used in the previous EQP implementation to form the velocity force vector.

rom/laghos_rom.cpp
```diff
@@ -338,6 +339,84 @@ void ComputeElementRowOfG_E(const IntegrationRule *ir,
...
+					hydrodynamics::QuadratureData const& quad_data,
+					Vector const& w_e, Vector const& v_e,
+					FiniteElement const& test_fe,
+					FiniteElement const& trial_fe,
+					const int zone_id,
+					const bool equationV, const bool equationE,
+					Vector & rv, Vector & re)
+{
+	const int nqp = ir->GetNPoints();
+	const int dim = trial_fe.GetDim(); // TODO: shouldn't it be the dimension of test_fe?
```
cval26, line 353: Shouldn't `dim` be the dimension of the velocity space `test_fe`, instead of the energy space `trial_fe`? In general, my understanding is that `test_fe` has dimension 3, while `trial_fe` has dimension 1. 

Variable `dim` is used to set the size of several vectors holding velocity coefficient values, so I'm not sure why the dimension of `trial_fe` is used.

rom/laghos_rom.cpp
cval26: I believe I resolved that in commit c4280af by adding the new option in `laghos_rom.hpp`.

rom/laghos_rom.cpp
```diff
@@ -338,6 +339,84 @@ void ComputeElementRowOfG_E(const IntegrationRule *ir,
...
+	Vector Vloc_force(loc_force.Data(), h1dofs_cnt * dim);
+
+	for (int q = 0; q < nqp; q++)
+	{
+		const IntegrationPoint &ip = ir->IntPoint(q);
+
+		// Form stress:grad_vshape at the current point.
+		// TODO: how do we set which velocity basis vector is used for the
+		// gradient calculation?
+		test_fe.CalcDShape(ip, grad_vshape);
```
cval26, line 380: Note my question in this part of the code. I'm not sure how change which basis vector is used in the gradient calculation.

rom/laghos_rom.cpp
```diff
@@ -2359,16 +2736,29 @@ ROM_Operator::ROM_Operator(ROM_Options const& input, ROM_Basis *b,
...
+	if (hyperreduce && hyperreductionSamplingType == eqp)
+	{
+		// basic EQP
+		// TODO: are reduced mass matrices needed for EQP, or just W matrices?
 
-        ComputeReducedMv();
-        ComputeReducedMe();
+		// TODO: why compute the reduced mass matrices, when for the basic EQP
+		// case we've already multiplied my M^{-1} and basis^T when deriving
+		// the reduced quadrature rules?
```
cval26, line 2746: Note my question here about computing the reduced mass matrices in the basic EQP case.

rom/laghos_rom.cpp
```diff
@@ -3027,6 +3417,7 @@ void ROM_Operator::UndoInducedGramSchmidt(const int var, Vector &S, bool keep_da
 
 void ROM_Operator::ApplyHyperreduction(Vector &S)
 {
+	// TODO: should set GramSchmidt = true when using energy-conserving EQP.
```
cval26, line 3420: where is that set?

rom/laghos_rom.cpp
```diff
@@ -4139,36 +4525,52 @@ void ROM_Operator::ForceIntegratorEQP(Vector & res) const
...
+			// energy-conserving EQP
+			for (int j = 0; j < rdim; ++j)
+			{
+				// TODO: v_e should be the jth velocity basis vector DOFs that
+				// correspond to the current element.
+				// Recall that the basis vectors have been renormalized by the
+				// induced Gram-Schmidt process, so that reduced Mv = I.
+
+				// I think the below is not right.
+				basis->GetBasisVectorV(false, j, v_e);
```
cval26: Here I want to set `v_e` to be the DOFs of the jth velocity basis vector corresponding to the current element, but I'm not sure how to do that. I started by getting the jth basis vector, but I believe I still need to reduce it to only the DOFs of the current element.

rom/laghos_rom.cpp
```diff
@@ -4388,36 +4799,41 @@ void ROM_Operator::StepRK2Avg(Vector &S, double &t, double &dt) const
         // - Update V using dv_dt.
         // - Compute de_dt and dx_dt using S and V.
 
-        // -- 1.
-        // S is S0.
+		// -- Stage 1 of 2 (S is S_n = S0).
         hydro_oper->UpdateMesh(fx);
```
cval26, line 4803: I haven't changed anything in the functionality of the average RK2 timestepper. I just added some comments that help me understand the carried out computations. I can remove them if you find them confusing.

rom/laghos_solver.cpp
```diff
@@ -277,7 +277,9 @@ void LagrangianHydroOperator::SolveVelocity(const Vector &S,
 
         if (use_eqp)
         {
-            rom_op->ForceIntegratorEQP_FOM(rhs);
+			// TODO: this function doesn't have access to the hyperreduction 
+			// sampling type; how should I resolve this?
+            rom_op->ForceIntegratorEQP_FOM(rhs, hyperreduction.SamplingType == eqp_energy);
```
cval26: In the current scope, I don't have access to the hyperreduction sampling type, which I want to pass as an argument so that the integrator function knows which RHS force vector to form. 

Where should I add the sampling type information so that it's within scope?

rom/laghos.cpp
```diff
@@ -394,8 +394,15 @@ int main(int argc, char *argv[])
...
-    MFEM_VERIFY(!(romOptions.SNS) || (romOptions.hyperreductionSamplingType != eqp), "Using SNS with EQP is prohibited");
+    romOptions.use_sample_mesh = romOptions.hyperreduce && (romOptions.hyperreductionSamplingType != eqp
+			|| romOptions.hyperreductionSamplingType != eqp_energy);
+
+    MFEM_VERIFY(!(romOptions.SNS) || (romOptions.hyperreductionSamplingType != eqp &&
+			romOptions.hyperreductionSamplingType != eqp_energy),
+			"Using SNS with EQP is prohibited");
+	
+	if (romOptions.hyperreduce && romOptions.hyperreductionSamplingType == eqp_energy)
+		romOptions.GramSchmidt = true;
```
cval26: I added this check here to ensure that Gram-Schmidt is performed when using the energy-conserving EQP. 

My understanding is that by setting this to true, we are normalizing the basis matrices so that the reduced mass matrices are identities. Is that correct?

rom/laghos_utils.cpp
```diff
@@ -507,11 +507,19 @@ void ReadPDweight(std::vector<double>& pd_weight, std::string outputPath)
...
     const int w = romOptions.window;
+
     romOptions.dimX = min(romOptions.max_dimX, twparam(w,0));
     romOptions.dimV = min(romOptions.max_dimV, twparam(w,1));
+
+	// TODO: for the energy-conserving EQP case, if twparam(w,2) exceeds the
+	// set maximum, truncating the stored basis will lead to loss of the
+	// energy identity vector, thus violating one of the energy conservation
+	// conditions. 
     romOptions.dimE = min(romOptions.max_dimE, twparam(w,2));
```
cval26, line 518: This function is called during the online timestepping stage, to set the window basis parameters for each time window. 

In our case, the last energy basis vector formed for each time window is the energy identity vector. If the dimension of the formed energy basis exceeds the set max value, do we truncate the stored basis to the new `dimE` set here? If we do so, then we'll remove the energy identity from the basis, thus violating the energy conservation principle.

rom/laghos_rom.cpp
```diff
@@ -986,6 +988,16 @@ void ROM_Sampler::SetupEQP_Force(const CAROM::Matrix* snapX,
...
 {
+	if (input.hyperreductionSamplingType == eqp_energy)
+	{
+		// For the energy-conserving EQP case, we increase the energy basis
+		// dimension by 1 to accomodate for the addition of the energy
+		// identity. 
+		// The change is done here, before the window basis parameters are
+		// written to their files, and is also carried to the caller's scope. 
+		cutoff[2] += 1;
+	}
```
cval26, line 999: I increase the energy snapshots cutoff value by 1 here, so that the correct value is written to the parameter files (to be read in the online stage).

rom/laghos_rom.cpp
```diff
@@ -1091,6 +1103,44 @@ void ROM_Sampler::Finalize(Array<int> &cutoff, ROM_Options& input)
...
+		CAROM::Matrix *tBasisE = basisE->getFirstNColumns(cutoff[2]);
+
+		// Form the energy identity and replace the last basis vector by it. 
+		Vector unitE(input.tL2Size);
+
+		// TODO: my understanding is that the "=" opetaror is overloaded, so
+		// that, by setting unitE = 1.0, the vector unitE contains the coeffs
+		// that result in the summed function being 1.0 on the grid points. 
+		// Is that correct?
+		unitE = 1.0;
```
cval26, line 1126: Note my question here: does that lead to the correct energy identity vector? 

I saw that this is the way the energy identity is also computed in a different part of the code; I just want to make sure this is the right way, and the right vector size.

rom/laghos_rom.cpp
```diff
@@ -1091,6 +1103,44 @@ void ROM_Sampler::Finalize(Array<int> &cutoff, ROM_Options& input)
...
+
+		for (int i = 0; i < input.tL2Size; i++)
+		{
+			// TODO: I believe I can access the last column in this way,
+			// without having to use pointer arithmetic, right?
+			(*tBasisE)(i, cutoff[2]-1) = unitE[i];
```
cval26, line 1132: I access the last column in this way, instead of relying directly on pointer arithmetic.

rom/laghos_rom.cpp
siuwuncheung: I think what we desire should be `F^T * v = (stress : grad v) phi_i`. There should be no velocity basis in the energy conservation equation.

rom/laghos_solver.cpp
siuwuncheung: The `ROM_Operator` class has a private variable `HyperreductionSamplingType hyperreductionSamplingType`. Perhaps you can write an public access function.

rom/laghos.cpp
siuwuncheung: That's correct. This will do the Gram Schmidt no matter what the user-defined option is. Although we may want to use Gram Schmidt for most of our runs, we may still want to keep the option alive while it exists. 

I'm a little concerned about whether the Gram Schmidt process is still functioning. We have been using SNS and not using Gram Schmidt for a while.

rom/laghos_rom.cpp
cval26: I believe that this is what I'm computing. My understanding is that `v_e * grad w_j` yields the gradient of the velocity snapshot we want, `grad v`. The reason I compute it this way is that it allows me to: 

1. use just one loop to calculate `Vloc_force` for both energy and velocity; 
2. most importantly, use the given `CalcDShape` function to get the gradient of the basis vector, `grad w_j`. Once I compute `grad w_j`, then I multiply it by the velocity snapshot coefficients, `v_e`, to get the desired `grad v`. Otherwise, I don't know how I can compute the gradient of the snapshot, `grad v`, directly. 

I believe that this is also the way the velocity force vector `F^T * v` is originally constructed in the paper by Dobrev et al (2012), where they outline the underlying FEM discretization.

siuwuncheung: Ah, I see. I overlooked the multiplication. I guess I am following the calculation now, and @dylan-copeland can further confirm it.

cval26: @dylan-copeland

rom/laghos_rom.cpp
cval26: @dylan-copeland

dylan-copeland: In the `else` clause below, `v_e` is set. It looks like you just need to copy and paste that code.

rom/laghos_rom.cpp
```diff
@@ -4647,35 +4660,45 @@ void ROM_Operator::ForceIntegratorEQP(Vector & res,
...
 		{
 			// energy-conserving EQP
 			for (int j = 0; j < rdim; ++j)
 			{
-				// TODO: v_e should be the jth velocity basis vector DOFs that
-				// correspond to the current element.
-				// I think the below is not right.
-				basis->GetBasisVectorV(false, j, v_e);	
+				// v_e: jth V basis vector's DOFs on this element
+				for (int k = 0; k < nvdof; ++k) v_e[k] = W_elems(eos + k, j);
```
cval26, line 4671: Here I access matrix `W_elems`, which has the data of `Wmat` organized by element, to get the V basis vector DOFs corresponding to the current element. 

This is similar to the way it's done in the basic EQP case (where, however, the basis vector is multiplied by Mv^{-1}) so that there's minimal changes to the code.

I do the calculation the same way for the energy equation.

rom/laghos_rom.cpp
```diff
@@ -3177,9 +3177,34 @@ void ROM_Operator::ComputeReducedMv()
                 (*Wmat)(i,j) = Mvj[i];
         }
     }
+	// TODO: do I need to enforce MPI rank == 0?
```
cval26, line 3181: Note my question here: do I need to do the computation of the reduced Mv inverse sequentially on one process only, or can it take place in parallel? For the basic EQP case is done in parallel, so I left it like that for this case as well.

rom/laghos.cpp
```diff
@@ -1581,8 +1581,7 @@ int main(int argc, char *argv[])
             romOper[0]->ApplyHyperreduction(romS);
         }
 
-		// To perform the induced Gram-Schmidt during the online stage, so that
-		// the reduced mass matrices are identity.
+		// TODO: do we want that for the energy-conserving EQP?
```
cval26, line 1584: Note my question here: do we want to call `ApplyHyperreduction` for the energy conserving EQP case? 

This function calls the induced Gram-Schmidt method, which we're not using anymore (since the Gram-Schmidt flag is now false). Then it calls a function `ComputeReducedMatrices` which I'm not sure about what it computes exactly. Among other things, I think it computes the matrix needed to get the RHS of the position equation (the Lagrangian part of the problem).

rom/laghos_rom.cpp
```diff
@@ -1105,6 +1095,22 @@ void ROM_Sampler::Finalize(Array<int> &cutoff, ROM_Options& input)
...
 	{
+		if (rank == 0)
+		{
+			// For the energy-conserving EQP case, we increase the energy basis
+			// dimension by 1 to accomodate for the addition of the energy
+			// identity. 
+			// The change is done here, before the window basis parameters are
+			// written to their files, and is also carried to the caller's scope. 
+			
+			// TODO: if cutoff[2] == windowNumSamples then increasing this by
```
cval26, line 1106: Should add a verification here until this gets fixed.

PR #178, opened 2023-07-06
Fix hypre link & build instructions

cval26:
Fixes the download link for `hypre` and updates the build instructions.


vladotomov (approved):
Thank you!

PR #184, opened 2024-02-23
Update for libROM PR 261

dylan-copeland:


cval26 (approved):

cval26:
Are the new libROM function interfaces backward compatible? Namely, if the time and timestep values are not removed from existing function calls, will that produce an error? Or have they been kept as default argument variables that are simply not used within libROM?

dylan-copeland:
> Are the new libROM function interfaces backward compatible? Namely, if the time and timestep values are not removed from existing function calls, will that produce an error? Or have they been kept as default argument variables that are simply not used within libROM?

The interface had a breaking change in libROM PR 261, i.e. it is not backward compatible. It is necessary to remove the time and timestep from the function calls, in order to compile. The changes are not too bad, as seen in this PR.

PR #198, opened 2025-10-27
shared device umpire pool

tomstitt:
Creates a single device Umpire pool for HYPRE and MFEM. This will make the device Umpire pool if laghos is built for the GPU and has Umpire even if the `device` is `"cpu"`; we need to create the Umpire pool before setting up MFEM/HYPRE. We could add parser for `device` in laghos or just not worry about it. thoughts?


v-dobrev (approved):
This makes sense to me. Thanks @tomstitt!

I didn't test it but it looks good.

PR #204, opened 2026-01-15
Add analytical Sedov shock solution

helloworld922:
Added a way to compare the Sedov density profile against the expected one by integrating \|\rho_{sim} - \rho_{exact}\|_2 over the domain.


helloworld922:
see #205 (different base branch)

PR #205, opened 2026-02-07
Add analytical Sedov shock solution

helloworld922:
Added a way to compare the Sedov density profile against the expected one by integrating |\rho_{sim} - \rho_{exact}|_2 over the domain.

PR #206, opened 2026-02-08
Better mesh generation + missing kernels

vladotomov:


laghos.cpp
```diff
@@ -268,47 +265,67 @@ int main(int argc, char *argv[])
...
    // On all processors, use the default builtin 1D/2D/3D mesh or read the
    // serial one given on the command line.
-   Mesh *mesh;
+   Mesh mesh;
+   Array<int> mpi_partitioning;
    if (strncmp(mesh_file, "default", 7) != 0)
    {
-      mesh = new Mesh(mesh_file, true, true);
+#ifndef MFEM_USE_METIS
+      MFEM_ABORT("MFEM has not been built with METIS. Use the "default" mesh.);
```
helloworld922:
```suggestion
      MFEM_ABORT("MFEM has not been built with METIS. Use the \"default\" mesh.");
```

laghos.cpp
```diff
@@ -268,47 +265,67 @@ int main(int argc, char *argv[])
...
+   TENS::Specialization<3,QVectorLayout::byNODES,1,1,2>::Opt<1>::Add();
+   TENS::Specialization<3,QVectorLayout::byVDIM,1,1,2>::Opt<1>::Add();
+   DET::Specialization<3,3,2,2>::Add();
+   GRAD::Specialization<3,QVectorLayout::byNODES,0,3,2,2>::Add();
+   GRAD::Specialization<3,QVectorLayout::byVDIM,0,3,2,2>::Add();
+   // 3D Q2Q1 - ok.
+   // 3D Q3Q2 - ok.
+   // 3D Q4Q3.
+   TENS::Specialization<3,QVectorLayout::byVDIM,3,5,8>::Opt<1>::Add();
+   DET::Specialization<3,3,5,8>::Add();
```
helloworld922: This exceeds the amount of shared memory on the GPU and fails to compile/link with CUDA or HIP.

laghos.cpp
```diff
@@ -268,47 +265,67 @@ int main(int argc, char *argv[])
...
+   // 2D Q1Q0.
+   TENS::Specialization<2,QVectorLayout::byNODES,1,1,2>::Opt<1>::Add();
+   TENS::Specialization<2,QVectorLayout::byVDIM,1,1,2>::Opt<1>::Add();
+   TENS::Specialization<2,QVectorLayout::byVDIM,2,2,2>::Opt<1>::Add();
+   GRAD::Specialization<2,QVectorLayout::byVDIM,0,2,2,2>::Add();
+   // 2D Q2Q1 - ok.
+   // 2D Q3Q2 - ok.
+   // 2D Q4Q3.
+   TENS::Specialization<2,QVectorLayout::byNODES,1,4,8>::Opt<1>::Add();
+   TENS::Specialization<2,QVectorLayout::byVDIM,2,5,8>::Opt<1>::Add();
```
helloworld922, line 286: requires https://github.com/mfem/mfem/pull/5227 for CUDA/HIP

laghos.cpp
```diff
@@ -268,47 +265,67 @@ int main(int argc, char *argv[])
...
+
+      // Read the serial mesh from the given mesh file on all processors.
+      // Refine the mesh in serial to increase the resolution.
+      mesh = Mesh::LoadFromFile(mesh_file, 1, 1);
+      for (int lev = 0; lev < rs_levels; lev++) { mesh.UniformRefinement(); }
    }
    else
    {
+      mesh = PartitionMPI(dim, Mpi::WorldSize(), elem_per_mpi, myid == 0,
+                          rp_levels, mpi_partitioning);
```
helloworld922, line 323: Do we want to extend this function to allow generating domains which are not the unit square/cube (similar to the Sx/Sy/Sz options I added to #205) so you can generate meshes for the triple point problem? I think this would also require specific multiples of elements in each dimension to ensure that interface boundaries match element boundaries.

This function also seems to generate some interesting elongated (in terms of zone count per dimension) meshes and only allows cartesian partitioning.

For 512 ranks in 3D I get 16x4x8, 1024 ranks gives 32x4x8. I think I prefer the existing mesh generation method as you can specify the exact domain shape and test metis partitioned mesh communication patterns/balance.

helloworld922 (changes requested):

laghos.cpp
vladotomov: @helloworld922 @camierjs  What are our options here?

Dropping Q4Q3 is an option, as I don't think they care about its performance too much. So if it's not a simple fix, we can just drop it.

laghos.cpp
vladotomov: As you said, that won't solve the problem for the 3point mesh. I prefer to keep it simple with `[0,1]^dim` - the intention is not to test physics, just a generic mesh that can be used for performance tests.

I will debug the weird elongations.

laghos.cpp
camierjs: We can drop it as it should only be used for the initial setup.

laghos.cpp
vladotomov: We can keep the Sx/Sy/Sz option + metis. I wanted to replace the complex `-partitionning` stuff.

laghos.cpp
helloworld922: I agree, just drop the specialization for DET Q4Q3.

vladotomov: done

laghos.cpp
vladotomov: @helloworld922 see
https://github.com/mfem/mfem/pull/5228

camierjs (approved):
Tested with https://github.com/mfem/mfem/pull/5227 - looks good!

helloworld922 (approved):

PR #209, opened 2026-03-18
Added configs to be part of Caliper metadata

amroakmal:


amroakmal:
@rfhaque : May you please review this PR? The goal is to add missing configs/params to Caliper metadata using Aidak. Please advise if there’re any changes needed.

laghos.cpp
```diff
@@ -139,6 +139,17 @@ int main(int argc, char *argv[])
...
    bool enable_nc = true;
 
+   #ifdef LAGHOS_USE_CALIPER
+      cali_config_set("CALI_CALIPER_ATTRIBUTE_DEFAULT_SCOPE", "process");
+      CALI_CXX_MARK_FUNCTION;
+   
+      MPI_Comm adiak_mpi_comm = MPI_COMM_WORLD;
+      void* adiak_mpi_comm_ptr = &adiak_mpi_comm;
+      adiak::init(adiak_mpi_comm_ptr);
+      adiak::launchdate();
```
daboehme: We should use `adiak::collect_all()` here to collect all the built-in metadata

pearce8 (changes requested):
Please address David's [comment](https://github.com/CEED/Laghos/pull/209/changes#r2963012530)

pearce8 (approved):

daboehme (approved):

vladotomov:
@amroakmal  Please resolve conflicts.
Also, 50 lines in the `main` function is too much, can you pull those into a separate function.

amroakmal:
@vladotomov : Thanks, I’ve resolved the conflicts and addressed the comments.

One point I wanted to clarify: I made these changes in main because, outside of it, I don’t have access to the variables needed to log the configuration values to Caliper. The current approach addresses your feedback, although it makes the main function more simpler/shorter and more readable, but the new function now takes several parameters.

I also considered an alternative approach of grouping these variables into a struct, but that would require more extensive changes than what I’ve done here.

Please let me know your thoughts and which approach you would prefer. Thanks.

vladotomov:
@amroakmal  this looks good, thank you.

PR #211, opened 2026-03-31
Add umpire pool size and memoy usage to caliper

rfhaque:
This PR:
- Adds umpire device memory pool size and the memory high water mark to caliper
- Prints umpire device memory pool size and the memory high water mark to the output log


rfhaque:
@vladotomov @pearce8 Please review and merge

pearce8 (approved):

PR #213, opened 2026-05-30
Bugfixes

helloworld922:
- fixed printing out backend before setting GPU-aware MPI
- fixed integer overflow for large number of global elements (causes h0 to be computed incorrectly)


vladotomov (approved):

kennyweiss (approved):
Thanks @helloworld922 !

PR #214, opened 2026-06-12
Enforce `--epm` exclusivity

michaelmckinsey1:
# Description
Simple change to enforce the exclusivity between `epm` and `nx`, `ny`, `nz`, `rs`, and `rp`. Currently, when setting `epm` these arguments will do nothing, but it would be clearer for users if Laghos raises error message when trying to use both.

# Example

```
$ laghos -p 1 -m default -epm 1048576 -nx 10

       __                __                 
      / /   ____  ____  / /_  ____  _____   
     / /   / __ `/ __ `/ __ \/ __ \/ ___/ 
    / /___/ /_/ / /_/ / / / / /_/ (__  )    
   /_____/\__,_/\__, /_/ /_/\____/____/  
               /____/                       

Option -epm/--elem-per-mpi cannot be used together with -nx/--xelems. Use -epm 0 to enable -nx/-ny/-nz/-rs/-rp.

Usage: laghos [options] 
...
```

### DIFF
--- a/laghos.cpp
+++ b/laghos.cpp
@@ -93,6 +93,8 @@ static void AssignMeshBdrAttrs3D(Mesh &, real_t, real_t, real_t, real_t);
 static long GetMaxRssMB();
 static void display_banner(std::ostream&);
 static void Checks(const int ti, const double norm, int &checks);
+static bool HasOption(int argc, char *argv[], const char *short_name,
+                      const char *long_name);
 
 #ifdef LAGHOS_USE_CALIPER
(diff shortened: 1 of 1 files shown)


laghos.cpp
```diff
@@ -273,6 +280,47 @@ int main(int argc, char *argv[])
       return 1;
    }
 
+   if (elem_per_mpi != 0)
```
vladotomov: Thanks, could you move these checks in a separate function?

michaelmckinsey1: Yes, done.

vladotomov (approved):

PR #215, opened 2026-07-27
Update CMake HIP build

scheibelp:
Use `LANGUAGE HIP` CMake 

Also add RAJA as a Laghos dependency

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ endif()
 find_package(HYPRE REQUIRED)
 find_package(MFEM REQUIRED)
 find_package(MPI REQUIRED)
+find_package(RAJA REQUIRED)
 
 option(LAGHOS_USE_CALIPER "Use Caliper" OFF)
 if (LAGHOS_USE_CALIPER)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ endif()
 find_package(HYPRE REQUIRED)
 find_package(MFEM REQUIRED)
 find_package(MPI REQUIRED)
+find_package(RAJA REQUIRED)
```
camierjs: This package should be made optional, only added by the user.

scheibelp: Thanks, this is addressed now

pearce8 (approved):

CMakeLists.txt
```diff
@@ -61,6 +66,13 @@ if (MFEM_USE_HIP)
   if(NOT DEFINED CMAKE_HIP_ARCHITECTURES)
     set(CMAKE_HIP_ARCHITECTURES gfx942)
   endif()
+  enable_language(HIP)
```
camierjs, line 69: Native HIP language support exists only from CMake 3.21, we might want to bump the global minimum to it.

camierjs (approved):

